PR #31, opened 2018-04-26
Capture tracebacks on slow method calls

itsderek23:
This adds the ability to capture tracebacks on slow method calls.


src/scout_apm/core/traceback.py
```diff
@@ -0,0 +1,21 @@
+# Module for helper functions to capture tracebacks
+
+import traceback
+
+
+def capture():
+    stack = traceback.extract_stack()
```
itsderek23: We may want to pass a `limit` parameter with a reasonable value. I [ran this script](https://gist.github.com/JettJones/c236494013f22723c1822126df944b12) to benchmark the performance impact of fetching the stack. This has a low limit value (2), so ours wouldn't be as fast. We need a deeper stack.

src/scout_apm/core/traceback.py
```diff
@@ -0,0 +1,21 @@
...
+    formatted_stack = []
```
itsderek23: Reviewing our [Ruby Backtrace Parser](https://github.com/scoutapp/scout_apm_ruby/blob/f2f84dd67d0a086fe52c0b0c49e0011315717502/lib/scout_apm/utils/backtrace_parser.rb) logic, it looks like we fetch 8 in-app frames (with a regex to determine if it is in-app).

Here's how an example looks right now for a Django ORM call in a view:

image

src/scout_apm/core/tracked_request.py
```diff
@@ -111,7 +116,15 @@ def tag(self, key, value):
...
+            return (self.end_time - self.start_time).total_seconds()
         else:
             # Current, running duration
-            (datetime.utcnow() - self.start_time).total_seconds()
+            return (datetime.utcnow() - self.start_time).total_seconds()
+
+    # Add any interesting annotations to the span. Assumes that we are in the
+    # process of stopping this span.
+    def annotate(self):
+        slow_threshold = 0.0
```
itsderek23: Our [Ruby logic for this](https://github.com/scoutapp/scout_apm_ruby/blob/c805e10520231390eaaadb8aad6258ba0aad6ca9/lib/scout_apm/tracked_request.rb#L146)...we set the threshold at 500 ms.

cschneid: Yep - set to 0 for testing purposes for now. 500ms or so is reasonable as a default.

PR #50, opened 2018-06-11
Add simple ContextManager and Decorator for custom instrumentation

cschneid:
An easy API to instrument specific spans of code:

```
import scout_apm.api

@scout_apm.api.instrument("Custom/MyCalculation")
def my_calc(self, x, y):
  pass

# or as a context manager:

def my_other_calc(self, x, y):
  with scout_apm.api.instrument("Custom/MyOtherCalculation") as instrument:
    instrument.tag("user_id", "1")
    pass
```

Questions:

* Should this be named `instrument`, or something more specific
* Should we limit the type prefix to be exactly "Custom"? If that string is high arity, it doesn't display well on the server. Elsewhere in the instrumentation it's "SQL" or "Template" or similar. Fixed strings.

### DIFF
--- a/src/scout_apm/api/__init__.py
+++ b/src/scout_apm/api/__init__.py
@@ -1,0 +1,43 @@
+from scout_apm.core.tracked_request import TrackedRequest
+
+import sys
+import logging
+
+
+# Logging
+logger = logging.getLogger(__name__)
+
+
+# Python 2 (and very early 3.x) didn't have ContextDecorator, so define it for ourselves
+if sys.version_info < (3, 2):
+    import functools
+
+    class ContextDecorator(object):
+        def __call__(self, f):
+            @functools.wraps(f)
+            def decorated(*args, **kwds):
+                with self:
+                    return f(*args, **kwds)
+            return decorated
+else:
+    from contextlib import ContextDecorator
+
+
+class instrument(ContextDecorator):
+    def __init__(self, operation):
+        self.operation = operation
+
+    def __enter__(self):
+        tr = TrackedRequest.instance()
+        self.span = tr.start_span(operation=self.operation)
+        return self
+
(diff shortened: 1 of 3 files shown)


cschneid:
A bit of docs for the finished version:

## API

Import the API module:

```python
import scout_apm.api

# or to not use the whole prefix on each call:

import instrument from scout_apm.api
```

`scout_apm.api.instrument(name, tags={}, kind="Custom")`

* `name` - A semi-detailed version of what the section of code is. It should be static between different invocations of the method. Individual details like user id, or counts or other data points can be added as tags. Names like `retreive_from_api` or `GET` are good names.
* `kind` - A high level area of the application. This defaults to `Custom`. Your whole application should have a very low number of unique strings here. In our built-in instruments, this is things like `Template` and `SQL`. For custom instrumentation, it can be strings like `MongoDB` or `HTTP` or similar. This should not change based on input or state of the application.
* `tags` - A dictionary of key/value pairs. Key should be a string, but value can be any json-able structure.

## As a Context Manager:

Wrap a section of code in a unique "span" of work.

The yielded object can be used to add additional tags individually.

```python
def foo():
  with scout_apm.api.instrument("Computation 1") as instrument:
    instrument.tag("record_count", 100)
    # Work

  with scout_apm.api.instrument("Computation 2", tags: { 'filtered_record_count: 50 } ) as instrument:  
    # Work
```

## As a decorator:

Wraps a whole function in a custom trace. Same API as the ContextManager style.

```python
@scout_apm.api.instrument("Computation", tags: { 'x': 1 })
def bar():
  # Work
```

itsderek23:
Help docs added: http://help.apm.scoutapp.com/#python-custom-instrumentation

Thanks @cschneid !

Issue #44, opened 2018-06-05
Exception `AttributeError: 'int' object has no attribute 'to_bytes'` 


bkvirendra:
Getting exception with Django 1.11.9 when using `scout_apm` integration, unable to get it started.

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/home/ubuntu/venv/local/lib/python2.7/site-packages/scout_apm/core/socket.py", line 80, in run
    self._register()
  File "/home/ubuntu/venv/local/lib/python2.7/site-packages/scout_apm/core/socket.py", line 157, in _register
    key=self.config.value('key')))
  File "/home/ubuntu/venv/local/lib/python2.7/site-packages/scout_apm/core/socket.py", line 128, in _send
    self.socket.sendall(self._message_length(data))
  File "/home/ubuntu/venv/local/lib/python2.7/site-packages/scout_apm/core/socket.py", line 143, in _message_length
    return length.to_bytes(4, 'big')
AttributeError: 'int' object has no attribute 'to_bytes'
```

My environment is as follows:
Python 2.7.12
Django 1.11.9
OS Ubuntu 16.04

cschneid:
Hi @bkvirendra we have a few known issues with python 2.7 keeping it from working. We intend to support it before coming out of tech preview, but for the moment it's broken. I'm sorry for the issue.

cschneid:
Closing this, since #56 fixed it. Keep track of the 2.7 support ticket (#45)

PR #56, opened 2018-06-12
Use `struct.pack` instead of int.to_bytes() for 2.7 support

cschneid:
Same outcome (a 4 byte big-endian number), but cross compatible with
python 2.7

Also fixes a typo in the unpack of received messages. All bytes over the
wire are big endian.


cschneid:
Will close #44

cschneid:
Table of magic pack/unpack strings: https://docs.python.org/2/library/struct.html

PR #61, opened 2018-06-19
Add additional thread locking around CoreAgentSocket

cschneid:
We were running into an issue (reproducable on 2.7, although likely
existing in other versions) where a socket thread was created, but then
not stored in the _instance variable. So it held the run_lock, but
wasn't available to be looked up.

This change adds a new lock around the instance call, which will prevent
race conditions around looking up, creating and recreating the socket
thread.

Also, it shifts the run_lock acquisition to the calling thread, instead
of being the first thing done in the newly created thread.


src/scout_apm/core/socket.py
```diff
@@ -23,33 +23,50 @@
...
-        return cls._instance
+        with cls._instance_lock:
+            if cls._instance is None:
+                cls._instance = cls(args, kwargs)
+                return cls._instance
+            elif cls._instance.running() is False:
+                cls._instance = cls(args, kwargs)
+                return cls._instance
+            else:
+                return cls._instance
```
dlanderson: remove the last else, and remove the first two returns, then just have one return at the end outside of the if/elif

src/scout_apm/core/socket.py
```diff
@@ -23,33 +23,50 @@
...
-            cls._instance = cls(args, kwargs)
-        elif cls._instance.running() is False:
-            del(cls._instance)
-            cls._instance = cls(args, kwargs)
-        return cls._instance
+        with cls._instance_lock:
+            if cls._instance is None:
+                cls._instance = cls(args, kwargs)
+                return cls._instance
+            elif cls._instance.running() is False:
```
dlanderson: I think this is wrong - we should just try to acquire the run lock instead. Also I don't think we need 3 events (starting/stopping/stopped).

dlanderson:
I don't know how I pulled in all these other commits. Closing this PR - see the actual changes in https://github.com/scoutapp/scout_apm_python/pull/92

PR #62, opened 2018-06-19
Add N+1 backtraces

dlanderson:


src/scout_apm/django/instruments/sql.py
```diff
@@ -29,15 +29,22 @@ def execute(self, sql, params=None):
...
+        span = tr.start_span(operation='SQL/Many')
         span.tag('db.statement', sql)
 
         try:
             return self.cursor.executemany(sql, param_list)
         finally:
-            TrackedRequest.instance().stop_span()
+            tr.stop_span()
+            tr.callset.update(sql, 1, span.duration_in_ms())
+            if tr.callset.capture_bracktrace(sql) is True:
```
cschneid: spelling.

src/scout_apm/core/call_set.py
```diff
@@ -0,0 +1,59 @@
...
+            return
+        self.call_count += call_count
+        self.call_duration += call_duration
+
+    def past_duration_threshold(self):
+        if self.past_duration_threshold is True:
+            # no need to check again once past
+            return True
+        self.past_time_threshold = self.call_duration >= self.__class__.DURATION_THRESHOLD
+        return self.past_duration_threshold
```
cschneid: is it actually faster to do the memoization of this vs. repeating a greater-than check?

src/scout_apm/core/call_set.py
```diff
@@ -0,0 +1,59 @@
+from __future__ import absolute_import
```
cschneid: Is there a better name than call_set?

`CallCounter` - `SpanCounter` - `SpanCounts` - `TracebackCapturePolicy` - `TracebackNPlusOnePolicy`

src/scout_apm/core/call_set.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class NPlusOneCallSet():
+    def __init__(self):
+        self.items = {}
+
+    def update(self, sql_string, call_count, call_duration):
+        self.find_for(sql_string).update(call_count, call_duration)
+
+    def capture_backtrace(self, sql_string):
```
cschneid: This method name is misleading because it does not capture a backtrace - it returns true/false for if a trace should be captured?

Also probably a missing `return` on next line.

src/scout_apm/core/call_set.py
dlanderson: I doubt it. Influenced from the existing Ruby check, but that used an actual DateTime object

src/scout_apm/django/instruments/sql.py
```diff
@@ -43,7 +43,7 @@ def executemany(self, sql, param_list):
         finally:
             tr.stop_span()
             tr.callset.update(sql, 1, span.duration_in_ms())
-            if tr.callset.capture_backtrace(sql) is True:
+            if tr.callset.should_capture_backtrace(sql) is True:
```
cschneid: the `is True` is redundant.

src/scout_apm/core/n_plus_one_call_set.py
```diff
@@ -0,0 +1,58 @@
...
+    DURATION_THRESHOLD = 150/1000.0
+
+    def __init__(self, sql_string):
+        self.sql_string = sql_string
+        self.captured = False
+        self.call_count = 0
+        self.call_duration = 0.0  # In Seconds
+        self.past_duration_threshold = False
+
+    def update(self, call_count, call_duration):
```
cschneid: Rather than 2 positional args, can we make this kwargs so it's not possible to flip them by accident?

PR #64, opened 2018-06-20
Initial Bottle support

dlanderson:


dlanderson:
To install:

In bottle app setup:

```python
from scout_apm.bottle import ScoutPlugin

app = bottle.default_app()
app.config.update({'scout.name': "YOUR_APP_NAME",
                                  'scout.monitor': "true",
                                  'scout.key': "YOUR_KEY"})

scout = ScoutPlugin()
bottle.install(scout)
```

pedropregueiro:
tested this locally on my server @dlanderson. it seems to work fine with the steps you added, but the traces seem quite poor, not sure if it's a problem with the Bottle integration or Scout itself.

Here's an example of an API call that goes through quite a few files: handlers, services, models, etc.
screenshot 2018-06-21 12 08 21

am I just looking at the wrong place here?

dlanderson:
We're getting our initial framework instrumentation done and then we're adding actual library instrumentation going (mongo, http calls, etc). Right now the bottle instrumentation only shows you the time spent handling the entire route request. Can you let us know what other libraries you're using or interested in having instrumented?

FYI, Scout for python is completely free during the tech preview (which is the next few months).

pedropregueiro:
ah cool, we're also using MongoDB (mongoengine) and Celery, which would be great to get some traces for.

do you know when will we be able to get some more data from the instrumentation, such as what classes and methods are called for each request and also, what time and % is spent there? just to have an idea of when we should give it another try

PR #65, opened 2018-06-21
Detect Git Sha from Env or Config

dlanderson:
Enable deploy tracking by detecting Git SHA via explicit config or Env vars

### DIFF
--- a/src/scout_apm/core/config/config.py
+++ b/src/scout_apm/core/config/config.py
@@ -4,6 +4,8 @@ import logging
 import os
 import platform
 
+from scout_apm.core.git_revision import GitRevision
+
 logger = logging.getLogger(__name__)
 
 
@@ -47,12 +49,16 @@ class ScoutConfig():
 
     def known_keys(self):
         return [
+            'app_server',
             'core_agent_dir',
             'core_agent_download',
             'core_agent_launch',
             'core_agent_version',
             'download_url',
+            'framework',
+            'framework_version',
             'git_sha',
+            'hostname',
             'log_level',
             'name',
             'key',
@@ -115,7 +121,6 @@ class ScoutConfigPython():
         return SCOUT_PYTHON_VALUES[key]
 
 
-
 class ScoutConfigEnv():
     """
     Reads configuration from environment by prefixing the key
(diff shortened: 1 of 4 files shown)


src/scout_apm/core/git_revision.py
```diff
@@ -0,0 +1,22 @@
...
+
+class GitRevision():
+    def detect(self):
+        sha = self.detect_from_env_var() or self.detect_from_heroku()
+        return sha or ''
+
+    def detect_from_heroku(self):
+        if 'HEROKU_SLUG_COMMIT' in os.environ:
+            return os.environ['HEROKU_SLUG_COMMIT']
+        return None
```
cschneid: Can simplify to be one-liner with `.get()`

```
>>> os.environ.get('MANPATH')
'/usr/share/man:/usr/local/share/man:/Users/cschneid/.fzf/man'
>>> os.environ.get('unknown')
>>> os.environ.get('unknown') is None
True
```

cschneid:
👍

src/scout_apm/core/config/config.py
```diff
@@ -153,13 +158,17 @@ def __init__(self):
...
+                'app_server': '',
                 'core_agent_dir': self.core_agent_dir,
                 'core_agent_download': True,
                 'core_agent_launch': True,
                 'core_agent_version': self.core_agent_version,
                 'download_url': 'https://s3-us-west-1.amazonaws.com/scout-public-downloads/apm_core_agent/release',
-                'git_sha': '',
+                'framework': '',
+                'framework_version': '',
+                'git_sha': GitRevision().detect(),
```
itsderek23: @cschneid - I believe we have a one-to-one match between config settings keys and env vars in all of our agents. This is nice. In this case, it looks like the config setting is `git_sha` and the env var is `SCOUT_REVISION_SHA`. 

I'd suggest we change the config to `revision_sha`. We use this naming for Ruby.

/cc @dlanderson

dlanderson: correct. We should use `revision_sha` for the setting. The `git_sha` is a field in `app_server_load` payloads

cschneid: FYI - this change was made after I merged this PR - d92227354fc3eeabf463c3bd01e667742d29c30b

PR #70, opened 2018-07-02
Add Web and BackgroundTransaction helpers in api package

cschneid:


src/scout_apm/api/__init__.py
```diff
@@ -44,3 +44,50 @@ def tag(self, key, value):
         if self.span is not None:
             self.span.tag(key, value)
 
+
+class Transaction(ContextDecorator):
+    def __init__(self, name, tags={}):
+        self.name = name
+        self.tags = tags
+
+    def start(kind, operation, tags={}):
```
dlanderson: Is this a `@classemthod`? Should be `start(cls, kind, operation, tags={})`?

src/scout_apm/api/__init__.py
```diff
@@ -44,3 +44,50 @@ def tag(self, key, value):
...
+        operation = kind + '/' + operation
+
+        tr = TrackedRequest.instance()
+        span = tr.start_span(operation=operation)
+        for key, value in tags.items():
+            tr.tag(key, value)
+        return span
+
+    def stop():
```
dlanderson: Also a classmethod?

src/scout_apm/api/__init__.py
```diff
@@ -44,3 +44,50 @@ def tag(self, key, value):
...
+        #  self.span = .start(self.name, self.tags)
+        #  return self.span
+
+    def __exit__(self, *exc):
+        return WebTransaction.stop()
+
+    def tag(self, key, value):
+        if self.span is not None:
+            self.span.tag(key, value)
+
```
dlanderson: Should we have a way to look up the current span?

src/scout_apm/api/__init__.py
```diff
@@ -44,3 +44,50 @@ def tag(self, key, value):
...
+        tr = TrackedRequest.instance()
+        span = tr.start_span(operation=operation)
+        for key, value in tags.items():
+            tr.tag(key, value)
+        return span
+
+    def stop():
+        tr = TrackedRequest.instance()
+        tr.stop_span()
+        return False
```
dlanderson: return `True` instead?

src/scout_apm/api/__init__.py
```diff
@@ -44,3 +44,50 @@ def tag(self, key, value):
...
+
+    #  def __enter__(self):
```
dlanderson: Not used?

src/scout_apm/api/__init__.py
```diff
@@ -44,3 +44,50 @@ def tag(self, key, value):
...
+        #  self.span = .start(self.name, self.tags)
+        #  return self.span
+
+    def __exit__(self, *exc):
```
dlanderson: What's `*exc`?

src/scout_apm/api/__init__.py
```diff
@@ -44,3 +44,50 @@ def tag(self, key, value):
         if self.span is not None:
             self.span.tag(key, value)
 
+
+class Transaction(ContextDecorator):
+    def __init__(self, name, tags={}):
+        self.name = name
```
dlanderson: It's a bit awkward to use `name` and `operation` interchangeably. I'd say we decide and use one or the other.

dlanderson: Also should we have a setter method to change the name/operation after the start, in cases where we don't know the name until further into the request? I don't know the guidelines on when to use setter method vs setting attribute directly.

src/scout_apm/api/__init__.py
cschneid: Shrug, I'll add it, but it's complexity that we aren't using.

src/scout_apm/api/__init__.py
cschneid: I suppose - it was false since the `__exit__` function needs `False` returned. But when called directly, I agree `True` makes more sense. 

https://docs.python.org/3/reference/datamodel.html#with-statement-context-managers

src/scout_apm/api/__init__.py
cschneid: In a `Controller/Foo` - the whole thing is operation, the `Foo` is name.

dlanderson: Then we should change the `operation` argument name in `start()` to `name` to be consistent, I think.

PR #90, opened 2018-07-25
Look up user_ip from the correct headers

cschneid:
Needs some testing still against a few real Django apps

### DIFF
--- a/src/scout_apm/django/middleware.py
+++ b/src/scout_apm/django/middleware.py
@@ -65,7 +65,7 @@ class ViewTimingMiddleware:
             if span is not None:
                 span.operation = 'Controller/' + view_name
                 Context.add('path', request.path)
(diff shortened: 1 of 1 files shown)


src/scout_apm/core/remote_ip.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @classmethod
+    def lookup_from_headers(cls, headers):
+        """
+        Given a dictionary of headers (WSGI request.META for instance), look up
+        the most likely user's IP
+        """
+
+        # A single address, set by this server, returned as an Array
+        remote_addr = RemoteIp.ips_from(headers.get("REMOTE_ADDR"))
```
dlanderson: Nitpick: might as well use `cls` instead of `RemoteIp`. If the class name ever changes you won't need to fix these references.

Issue #91, opened 2018-07-25
 AttributeError: type object 'CoreAgentSocket' has no attribute '_instanceAttribute'


jobelenus:
AttributeError: type object 'CoreAgentSocket' has no attribute '_instance'
  File "django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "scout_apm/django/middleware.py", line 55, in __call__
    tr.stop_span()
  File "scout_apm/core/tracked_request.py", line 73, in stop_span
    self.finish()
  File "scout_apm/core/tracked_request.py", line 86, in finish
    RequestManager.instance().add_request(self)
  File "scout_apm/core/request_manager.py", line 20, in add_request
    self.request_buffer.add_request(request)
  File "scout_apm/core/request_manager.py", line 30, in add_request
    self.flush()
  File "scout_apm/core/request_manager.py", line 36, in flush
    self.flush_request(request)
  File "scout_apm/core/request_manager.py", line 42, in flush_request
    AgentContext.socket().send(batch_command)
  File "scout_apm/core/context.py", line 18, in socket
    return CoreAgentSocket.instance(scout_config=ScoutConfig())
  File "scout_apm/core/socket.py", line 30, in instance
    if cls._instance is None:

This happens with a POST from mailgun to a django view. That view in particular has two decorators:
```
@csrf_exempt
@transaction.non_atomic_requests
def mailgun_store_webhook(request):
```

This seems to be intermittent (e.g. I've seen two failures so far, but other webhook POSTs are succeeding in my application.

python==3.6.4
Django==1.11.14
scout-apm==1.1.8

cschneid:
Thank you for the report - I'll take a look at this.  Do you have any sort of async or other multi-threaded  code serving requests?

jobelenus:
We are serving with daphne==2.2.1, on Django channels==2.1.2, but this is a normal (sync) Django view, not a websocket response (either sync or async)

dlanderson:
Will be released in v1.2.2 next week

dlanderson:
Resolved by https://github.com/scoutapp/scout_apm_python/pull/92

PR #92, opened 2018-07-26
Adds a lock around CoreAgentSocket.instance(), reworks some condition…

dlanderson:
…s for races


dlanderson:
Resolves https://github.com/scoutapp/scout_apm_python/issues/91

PR #99, opened 2018-07-30
Remove "Flask" span in Flask instrumentation

dlanderson:


cschneid:
I talked with @dlanderson about this previously - I added the Flask span initially because I could, rather than for any particular instrumentation reason. From what we've seen, it only times the internals of Flask, which are uninteresting, and unlikely to be user related.

👍

PR #103, opened 2018-08-07
Initial objtrace extension

dlanderson:
We suspect Python suffers from the same performance overhead with object allocation as Ruby. This PR creates a `scout_apm.objtrace` module that hooks into Python's memory allocator and counts the number of Python objects being allocated.

Before `scout_apm` is installed, we try to compile the C extension ourselves. If the compilation fails, `scout_apm` will still install without but it will not record the object allocations.

Not yet ready to merge. Needs to wire up the collection and reporting of the allocations into the `TrackedRequests`


dlanderson:
Also adds `mem_delta` tag to TrackedRequest. This is the increase of RSS size in MB during the request.

PR #106, opened 2018-08-13
Extract a RemoteIp library and use it in Django Middleware

cschneid:


cschneid:
Mistaken pull request, will reopen.

Issue #104, opened 2018-08-10
Error with init_app in flask app


andrelramos:
I'm getting a error in my application when i use the init_app method to start the scout on my flask app. Follow the bellow example:

```python
scoutapm = ScoutApm()

def create_app():
	app = Flask(__name__)
	scoutapm.init_app(app)
```

This code generate the bellow exception:
```jsx
Traceback (most recent call last):
  File "/home/andreramos/.virtualenvs/bandroid/lib/python3.6/site-packages/werkzeug/serving.py", line 270, in run_wsgi
    execute(self.server.app)
  File "/home/andreramos/.virtualenvs/bandroid/lib/python3.6/site-packages/werkzeug/serving.py", line 258, in execute
    application_iter = app(environ, start_response)
  File "/home/andreramos/.virtualenvs/bandroid/lib/python3.6/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/andreramos/.virtualenvs/bandroid/lib/python3.6/site-packages/raven/middleware.py", line 100, in __call__
    iterable = self.application(environ, start_response)
  File "/home/andreramos/.virtualenvs/bandroid/lib/python3.6/site-packages/flask/app.py", line 1985, in wsgi_app
    response = self.handle_exception(e)
  File "/home/andreramos/.virtualenvs/bandroid/lib/python3.6/site-packages/flask_restful/__init__.py", line 273, in error_router
    return original_handler(e)
  File "/home/andreramos/.virtualenvs/bandroid/lib/python3.6/site-packages/flask_restful/__init__.py", line 273, in error_router
    return original_handler(e)
  File "/home/andreramos/.virtualenvs/bandroid/lib/python3.6/site-packages/flask_restful/__init__.py", line 273, in error_router
    return original_handler(e)
  File "/home/andreramos/.virtualenvs/bandroid/lib/python3.6/site-packages/flask_restful/__init__.py", line 273, in error_router
    return original_handler(e)
  File "/home/andreramos/.virtualenvs/bandroid/lib/python3.6/site-packages/flask_restful/__init__.py", line 273, in error_router
    return original_handler(e)
  File "/home/andreramos/.virtualenvs/bandroid/lib/python3.6/site-packages/flask_restful/__init__.py", line 273, in error_router
    return original_handler(e)
[...]

PR #107, opened 2018-08-13
Save App variable in Flask instruments `init_app(app)`

cschneid:
A customer had a workflow where the ScoutApm() object was created before
the Flask app, and then the app was passed to `init_app`. Because we
didn't save it off at that point, it caused an exception when accessing
`self.app`.

```
scoutapm = ScoutApm()

def create_app():
  app = Flask(__name__)
  scoutapm.init_app(app)
```


cschneid:
See #104

PR #113, opened 2018-08-16
Fix RemoteIp class looking for the wrong keys in the header array

cschneid:
The header dictionary created by WSGI has keys with a prefix of `HTTP_`

But not for REMOTE_ADDR. Since that's not a user-provided value, and
instead generated by the server - it's not a header, but an extra bit of
data that lives with the headers


dlanderson:
👍

PR #122, opened 2018-08-27
Reverse Django middleware insertion logic

dlanderson:
Fixes a bug in Django <= 2 when both `MIDDLEWARE` and `MIDDLEWARE_CLASSES` are defined.


cschneid:
Situation across versions:

1.8/1.9 - MIDDLEWARE_CLASSES only
1.10/1.11 - Both
2.0 - MIDDLEWARE only.

Issue #125, opened 2018-08-30
CoreAgentSocket exception on _send: timeout('timed out',)


itsderek23:
With [this demo app](https://github.com/itsderek23/bakerydemo), I see the following in the log after every-other-request I trigger in the browser:

```
CoreAgentSocket exception on _send: timeout('timed out',)
```

That said, it appears that every request I trigger does appear in the Scout UI. This is the log as I trigger 2 requests:

* [Python agent log](https://gist.github.com/itsderek23/9cb6b28faa4e2abad573812c865c14b8)
* [Core Agent](https://gist.github.com/itsderek23/be45ddb7ccdb65f5cd374373064c97dc)

To setup the app, I followed [this set of instructions](https://github.com/itsderek23/bakerydemo#setup-with-virtualenv).

As specified, I run the app via `./manage.py runserver`.

I setup a `.env` file with:

```
DJANGO_SETTINGS_MODULE=bakerydemo.settings.local
SCOUT_KEY=KEY
SCOUT_LOG_FILE=core.log 
SCOUT_LOG_LEVEL=trace
``` 

As an easy first step, it might be worthwhile testing this against any Django app with `./manage.py runserver` running with `DEBUG = True`.

PR #126, opened 2018-09-04
Always read the response to CoreAgent messages

cschneid:
This appears to fix the socket timeout errors in #125.

Previously we were simply ignoring the contents of any response message
from the core agent, not even reading off the socket. We were seeing
`timed out` exceptions. By reading off the socket every time, and fixing
the `recv` calls to actually read *all* of the data off the socket, the
timeouts go away on my system.

### DIFF
--- a/src/scout_apm/core/socket.py
+++ b/src/scout_apm/core/socket.py
@@ -6,6 +6,7 @@ import socket
 import struct
 import time
 import threading
+import os
 
 try:
     # Python 3.x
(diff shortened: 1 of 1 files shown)


src/scout_apm/core/socket.py
```diff
@@ -145,16 +146,21 @@ def _send(self, command, do_async=True):
...
-            logger.debug("CoreAgentSocket exception on _send: %s" % repr(e))
+            logger.debug("CoreAgentSocket exception on data _send: %s on PID: %s on thread: %s" % (repr(e), os.getpid(), threading.current_thread()))
             return None
 
         if do_async is True:
             return True
         else:
             # TODO read respnse back in to command
-            self._read_response()
+            response = self._read_response()
```
dlanderson: Save an allocation by removing `response =`

cschneid: You're right it's an unused variable, but I'm going to leave it for clarity when we start needing the responses from the core agent. It doesn't hurt anything right here.

PR #127, opened 2018-09-10
Add a lock around ThreadLocalSingleton tasks

cschneid:
We've had customer reports of races in this class when run with python
async code, which uses finer-grained multithreading than actual threads.

Adding the lock here should fix this. It worked in socket.py.

### DIFF
--- a/src/scout_apm/core/thread_local.py
+++ b/src/scout_apm/core/thread_local.py
@@ -4,23 +4,37 @@ import threading
 
 
 class ThreadLocalSingleton(object):
+    _instance_lock = threading.Lock()
+
     @classmethod
     def instance(cls, *args, **kwargs):
-        if not hasattr(cls, '_thread_lookup'):
-            cls.__new_instance(args, kwargs)
-        elif not hasattr(cls._thread_lookup, 'instance'):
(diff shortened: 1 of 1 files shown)


src/scout_apm/core/thread_local.py
```diff
@@ -4,23 +4,37 @@
...
+
+            # Somehow we have the thread local dict, but failed to make the instance. Just reinit both
+            if not hasattr(cls._thread_lookup, 'instance'):
+                cls.__new_instance(args, kwargs)
+                return cls._thread_lookup.instance
+
+            # The instance is none, so we'll need to recreate it (and the thread local dict too)
+            if cls._thread_lookup.instance is None:
+                cls.__new_instance(args, kwargs)
+                cls._thread_lookup.instance
```
dlanderson: What's the purpose here?

src/scout_apm/core/thread_local.py
```diff
@@ -4,23 +4,37 @@
...
+
+            return cls._thread_lookup.instance
```
dlanderson: Either be consistent and return in every `if` case, and log an error if you reach here. Or don't return in the `if` scopes and instead only one return at the end.

PR #129, opened 2018-10-17
Rename transaction

cschneid:
This matches the Ruby agent's ScoutApm::Transaction.rename call.

This is most useful when an app has a single entry point, then makes business
logic decisions with easily nameable branches. GraphQL for instance has a
single POST endpoint, but immediately branches into different queries and
mutations with clear names.

**NOTE** - This requires a yet-unreleased change to the binary agent, and will
not yet work. A note will be added to this PR when it is fully released


dlanderson:
@cschneid Is this supported in core agent yet?

cschneid:
It isn't, I'll add it to the list of next version things. Created https://github.com/scoutapp/core-agent/issues/69 to track.

adamchainz:
👍

tests/scout_apm_tests/api/rename_transaction_test.py
```diff
@@ -0,0 +1,12 @@
+import scout_apm.api
```
adamchainz: I think the test should be in `integration/test_api.py` with all the other tests for the public intefrace

tests/scout_apm_tests/api/rename_transaction_test.py
```diff
@@ -0,0 +1,12 @@
...
+from scout_apm.core.tracked_request import TrackedRequest
+
+
+def test_rename_transaction():
+    tr = TrackedRequest.instance()
+    assert('transaction_name' not in tr.tags)
```
adamchainz: no brackets, though `black` will sort that for you, also shouldn't that be `transaction.name` with a dot?

cschneid: yes, it should be

PR #132, opened 2018-11-17
Support Python 2.7 - phase 1 - cleanup

aaugustin:
This PRs normalizes the code base according to Python best practices and lays groundwork for supporting Python 2.7. It doesn't introduce any behavior changes on Python 3.4+.

It consists in:

- [x] Updating the target Python (and Django) versions
- [x] Normalizing imports with isort (and some manual work)
- [x] Normalizing code formatting with black
- [x] Fixing flake8 warnings and removing as many exceptions as possible.

as well as some smaller cleanups.

I took care of making independent, small commits (or automated refactorings) to make it easier to review and confirm that these cosmetic changes are harmless.

I'm proposing to merge it into master because I think it's a strict improvement.


aaugustin:
This PR is still a work in progress. I'm going to rebase it heavily. I'll say when it's ready for review.

aaugustin:
Tips for reviewing this pull request:

- Hopefully most commits are small and self-explanatory.
- Don't bother reading "Run isort." and "Run black." — changes are very large but fully automated. isort and black are reliable. black even checks that Python produces the same AST before and after reformatting.
- Please read "Fix all flake8 warnings." and "Format logging strings lazily." — they shouldn't change anything and I proofread them several times but human error is always a possibility when making many small changes.

setup.py
```diff
@@ -1,65 +1,72 @@
...
+)
+if os.path.exists("README.md"):
+    long_description = open("README.md").read()
 
 # Try to compile the extensions, except for platforms or versions
 # where our extensions are not supported
 compile_extensions = True
 
 setup_args = {
-      'name': 'scout_apm',
```
dlanderson: I explicitly used single quotes unless a string needs interpolation. Ultimately it probably doesn't matter, but are we switching to double quotes just because of historical preference?

setup.py
```diff
@@ -1,65 +1,72 @@
...
+if sys.platform.startswith("java"):
     compile_extensions = False
 
-if '__pypy__' in sys.builtin_module_names:
+if "__pypy__" in sys.builtin_module_names:
     compile_extensions = False
 
-if compile_extensions is not True:
-    del setup_args['ext_modules']
+if not compile_extensions:
```
dlanderson: These comparisons are not equivalent. `if compile_extensions is not True` is looking for exactly `True` and not just any truthy value. I generally prefer the explicit `is not True` in order to prevent bugs where we have some truthy value unintentionally.

src/scout_apm/bottle/__init__.py
```diff
@@ -1,26 +1,25 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, division, print_function, unicode_literals
 
 from bottle import request
 
 import scout_apm.core
-from scout_apm.core.context import AgentContext
+from scout_apm.api.context import Context
```
dlanderson: These are not the same. Unfortunately we use the word `Context` in both modules.

`AgentContext` is for `scout_apm` internal/core use only and provides a `state of the agent`.

`Context` provides a way for our instrumentation and end-users to attach bits of information to a `TrackedRequest`

dlanderson: Oops, I see this was just an ordering change. It was confusing to see in the GitHub diff. Ignore.

src/scout_apm/bottle/__init__.py
```diff
@@ -30,42 +29,43 @@ def setup(self, app):
         scout_apm.core.install()
 
     def apply(self, callback, context):
-        if AgentContext.instance.config.value('monitor') is not True:
+        if not AgentContext.instance.config.value("monitor"):
```
dlanderson: Another truthiness check change as commented above

src/scout_apm/core/cli/core_agent_manager.py
```diff
@@ -1,39 +1,40 @@
-#!/usr/bin/env python
```
dlanderson: Isn't this needed if a user wants to execute this file from a shell?

src/scout_apm/core/core_agent_manager.py
```diff
@@ -9,41 +8,45 @@
...
+        self.downloader = CoreAgentDownloader(
+            self.core_agent_dir,
+            AgentContext.instance.config.value("core_agent_full_name"),
+        )
 
     def launch(self):
-        if AgentContext.instance.config.value('core_agent_launch') is not True:
-            logger.debug("Not attempting to launch Core Agent "
-                         "due to 'core_agent_launch' setting.")
+        if not AgentContext.instance.config.value("core_agent_launch"):
```
dlanderson: I'm not going to mark all truthiness changes - just marking this one again to ensure we discuss them as a whole.

src/scout_apm/core/core_agent_manager.py
```diff
@@ -53,77 +56,79 @@ def download(self):
...
-                    )
-            process.wait()
+            subprocess.check_call(
+                self.agent_binary()
+                + self.daemonize_flag()
+                + self.log_level()
+                + self.log_file()
+                + self.config_file()
+                + self.socket_path()
+            )
```
dlanderson: It probably does, but just verifying this is the same behavior as `Popen`+`wait`

tox.ini
```diff
@@ -26,8 +29,9 @@ deps =
     pytest-travis-fold
     pytest-cov
     django18: Django>=1.8,<1.9
-    django19: Django>=1.9,<2.0
+    django111: Django>=1.11,<2.0
```
cschneid: I guess the removed line resolved to 1.11 anyway, so this is clearer?

setup.py
aaugustin: There are two common conventions in Python-land:

A. Single quotes for "technical strings" (~ Ruby symbols) and double quotes for "human-readable strings" (~ could be translated)
B. Double quotes for everything (except perhaps strings containing double quotes, in order to avoid backslash-escaping)

By default `black` uses option B. I used to be a fan of option A; `black` made me realize it creates mental overhead for almost no practical benefits. I'm now a convert to option B aka "just don't think about it".

I can re-run `black` with the `-s` option which preserves the choice of single or double quotes if you'd like (and rebase the rest of the pull request, which is non-trivial work because there will be lots of conflicts).

Conclusion: just tell me if this convinces you to go with double quotes or if you'd like me to preserve the original quote style.

cschneid: I don't care one bit, beyond not creating too much PR noise due to disagreement about it. Do the change here, and it looks like unit tests will enforce this style going forward?

src/scout_apm/core/cli/core_agent_manager.py
aaugustin: Yes, but executing this file from a shell doesn't do anything (as it currently stands).

setup.py
dlanderson: Ha, I know you know the difference :)

I suppose there are quite a few things I don't agree with in Python, this being one of them. I don't mind using `if not compile_extensions` style to check for existence/presence, but I dislike it when I'm actually checking for a specific value. If it's a code smell then we can change it as you have.

`and doesn't cause bugs.` 🤞Let's hope! We've seen it in Ruby ;)

setup.py
aaugustin: Yes, `make style` will auto-format the code and `tox -e check` will enforce the auto-formatting going forwards.

src/scout_apm/core/core_agent_manager.py
aaugustin: Oops — good catch, actually it isn't the same behavior if the process exits with a return code other than 0.

Before: `wait()` returns that code, which is ignored: `run()` returns `True`.

After: `check_call()` raises `CalledProcessError`, which is caught; `run()` returns `False`.

Which is the correct behavior? (Really it depends whether an exit code other than 0 signals a failure.)

Perhaps this change took care of `# TODO detect failure of launch properly`!

setup.py
```diff
@@ -1,65 +1,72 @@
...
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+    ],
+}
 
 if sys.version_info <= (3, 0):
     compile_extensions = False
 
-if sys.platform.startswith('java'):
+if sys.platform.startswith("java"):
```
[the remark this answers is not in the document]
aaugustin: Actually I'm not sure about your linting setup. I saw a handful of `pylint` exclusions but no references to `pylint` in `tox.ini` or elsewhere.

I think isort + black + flake8 is the sweet spot these days.

Obviously I can adapt to whatever you like best.

src/scout_apm/core/__init__.py
```diff
@@ -1,6 +1,5 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, division, print_function, unicode_literals
```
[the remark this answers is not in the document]
aaugustin: It's every `__future__` import that isn't enabled by default in Python 2.7 and is in Python 3.

Essentially it's "make Python 2.7 behave as much as possible like Python 3".

It's much easier to always include the whole list than to check and update it whenever you write a division or a `print()`.

src/scout_apm/core/config/__init__.py
```diff
@@ -40,33 +42,30 @@ def locate_layer_for_key(self, key):
...
     def log(self):
-        logger.debug('Configuration Loaded:')
+        logger.debug("Configuration Loaded:")
         for key in self.known_keys():
             layer = self.locate_layer_for_key(key)
-            logger.debug('{:9}: {} = {}'.format(
-                layer.name(),
-                key,
-                layer.value(key)))
+            logger.debug("%-9s: %s = %s", layer.name(), key, layer.value(key))
```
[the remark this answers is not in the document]
aaugustin: No, it's quite old. It works in 2.7 (probably in all 2.x versions and perhaps even in 1.x versions).

```
>>> "%-9s" % "foo"
'foo      '
>>> "%9s" % "foo"
'      foo'
```

tox.ini
aaugustin: 1.8 and 1.11 are LTS versions.

I thought there was little benefit to testing obsolete, unsupported, not LTS versions. That's really why I skipped 1.9 and 1.10.

However, we may be hooking into APIs that were refactored around these versions, so it may be worth adding them.

src/scout_apm/core/cli/core_agent_manager.py
dlanderson: Ah, right. I forgot the `entry_points` spec handles setting up CLI scripts.

src/scout_apm/core/core_agent_manager.py
dlanderson: A check (just to log failure) might be good. The goal is to have python try to start it (99% of cases it will), and if launch fails then we continue as normal. If there is no core agent running then we don't interrupt the python process at all but we will see log messages about not being able to communicate with it.

setup.py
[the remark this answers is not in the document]
dlanderson: No preference. Let's go with what you have, Aymeric.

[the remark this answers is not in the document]
cschneid: Same, no preference, I had my editor enforcing pylint, but not in the build / test sequence. This is good

src/scout_apm/core/core_agent_manager.py
aaugustin: Yes, the code does this. When launching the core agent fails and `check_call()` raises an exception, it's caught and logged:
```python
logger.error("Error running Core Agent: %r", e)
```

tox.ini
dlanderson: I'm vaguely remembering some changes between 1.8 and 1.9 but I can't remember what. Perhaps @cschneid remembers.

cschneid: I don't remember any specifics. These 2 versions are fine for the moment.

There's a certain level of support we have to have even for non supported versions, as long as enough customers require it. But I bet 1.8 and 1.11 tests will catch 99% of code that may have changed between the two.

PR #135, opened 2018-11-24
Improve and simplify Travis CI setup.

aaugustin:
This is ready for review.

It rewrites significantly the Travis CI integration. The main benefits are:
- the configuration is self-contained; it only depends on tox
- stages make it more obvious that tests run first (still not super obvious because Travis CI runs them  implicitly), then, only if they succeed, artifacts are built
- tests run on Python 2.7
- style checks run
- tests of wheels actually use the version installed from a wheel (I'm not convinced it's super useful, but since it was implemented, I fixed it)

According tho the [Travis CI docs](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings), it would be best to set the TWINE_USERNAME and TWINE_PASSWORD environment variables in the [Travis CI settings](https://travis-ci.org/scoutapp/scout_apm_python/settings) and remove them from `.travis.yml`. I can't do that myself because I don't know the value of TWINE_PASSWORD.


aaugustin:
Adding `codecov` would be a good follow-up. I requested permissions to do it.

PR #137, opened 2018-11-25
Add a WSGI application for manual testing.

aaugustin:
Add a WSGI application for manual testing.


aaugustin:
To generate traffic with this app: `while true; do sleep 1; wget --mirror http://localhost:8080/; done`

PR #143, opened 2018-12-18
Rename a Transaction

cschneid:
This matches the Ruby agent's ScoutApm::Transaction.rename call.

This is most useful when an app has a single entry point, then makes business
logic decisions with easily nameable branches. GraphQL for instance has a
single POST endpoint, but immediately branches into different queries and
mutations with clear names.

**NOTE** - This requires a yet-unreleased change to the binary agent, and will
not yet work. A note will be added to this PR when it is fully released

### DIFF
--- a/src/scout_apm/core/config.py
+++ b/src/scout_apm/core/config.py
@@ -64,6 +64,7 @@ class ScoutConfig(object):
             "framework",
             "framework_version",
             "hostname",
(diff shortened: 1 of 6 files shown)


cschneid:
The `hub` command put another branch's description in. Closing, will reopen in a bit.

PR #144, opened 2018-12-18
New setting `ignore` to ignore path prefixes

cschneid:
Allows you to easily ignore /health and similar high-throughput, low
meaning endpoints. Any requests to paths that start with any of the
ignore list will not be sent, and hence not counted as Throughput, and
not averaged into average or 95th percentile response times, and no
traces will be captured.


dlanderson:
Released in v2.0.0

PR #146, opened 2019-01-07
Remove all references to PyYAML

### LINKED ISSUE #145: PyYAML security alert
Our application is getting a GitHub security alert from PyYAML, and this is the only package that has it as a dependency in our graph. It looks like this package no longer uses that package, but it is still declared as a dependency.

If this assessment is correct, the dependency should be removed from the `setup.py` and a new release upload to PyPI.

https://nvd.nist.gov/vuln/detail/CVE-2017-18342

cschneid:
A user reported that an automated GitHub security scan reported PyYAML
as being vulnerable [1].  The user also said that it doesn't appear that
we even use PyYAML. After some checking, it looks that he's right, we
don't use it.

My memory of how this dependency was ever added is that we pulled it in
as a dependency very early in development to implement configuration in
a similar way to how the Ruby agent does. But then, we implemented a
more Pythonic system for configuration.

The underlying security issue isn't relevant to or exploitable in our
agent. We never even import the PyYAML into our code, much less use the
potentially vulnerable `load` function. But of course, the most secure
code is code that doesn't exist. So this commit removes all mentions of
'pyyaml'.

[Closes #145]

[1]: https://github.com/scoutapp/scout_apm_python/issues/145

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -89,7 +89,7 @@ for example 2.7 and 3.7. In that case, do the whole setup for each virtualenv.
 
 Install all test dependencies:
 
-    $ pip install bottle celery Django elasticsearch flask flask-sqlalchemy jinja2 mock psutil pymongo pyramid pytest pytest-travis-fold pytest-cov PyYAML redis requests sqlalchemy urllib3 webtest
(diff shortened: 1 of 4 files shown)


cschneid:
@dlanderson Can you review and doublecheck that yaml is indeed an unused dependency?

PR #153, opened 2019-03-04
Tracking tr

xiamx:


cschneid:
Hi @xiamx - can you describe the change and use of what you're working on? We can help if there's a specific use case you're trying to implement.

xiamx:
@cschneid I created this cross fork PR by accident, sorry, was trying to create a PR within my branch. I'm testing out on somethings to make scoutapm python work with grpc, I'll create a real PR when I'm ready

cschneid:
Let me know what I can do to help. Watch out for changing public APIs, we want to avoid breaking public contracts if there's any way around it.

PR #154, opened 2019-03-06
Fixes "Registering with app" log message not using configured logger

vishalmanohar:
The log message was defaulting to use root logger config and not the logger configured for scoutapp


cschneid:
Great catch! Thank you.

PR #156, opened 2019-03-29
Fix Elasticsearch integration for 'index' being passed as a list

adamchainz:
`elasticsearch-py` documented each method as taking the `index` parameter as a string (either a single index name or a comma separated list). The higher level library `elasticsearch-dsl-py` [forces using a list of strings](https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/search.py#L100) for the `index` parameter that it passes down to `elasticsearch-py`, which is supported but not documented as it is later serialized to a comma-separated list (in [`_make_path`](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L44)).

Our monkey patch on all the methods in `elasticsearch-py` only expected strings, this updates it to work with lists of strings as well. I added a test that failed before and passes after.

I've also made [a PR to `elasticsearch-py` to fix their documentation](https://github.com/elastic/elasticsearch-py/pull/921).

### DIFF
--- a/src/scout_apm/instruments/elasticsearch.py
+++ b/src/scout_apm/instruments/elasticsearch.py
@@ -84,11 +84,14 @@ class Instrument(object):
 
         for method_str in self.__class__.CLIENT_METHODS:
             try:
-                code_str = """
+                code_str = """\
 @monkeypatch_method(Elasticsearch)
 def {method_str}(original, self, *args, **kwargs):
     tr = TrackedRequest.instance()
-    index = kwargs.get('index', 'Unknown').title()
+    index = kwargs.get('index', 'Unknown')
+    if isinstance(index, (list, tuple)):
+        index = ','.join(index)
+    index = index.title()
     name = '/'.join(['Elasticsearch', index, '{camel_name}'])
     tr.start_span(operation=name, ignore_children=True)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Pending Release
+
+### Fixed
+
+- Fixed Elasticsearch integration for queries passing 'index' to `elasticsearch-py` as a list (PR #156)
+
 ## [2.0.1] 2019-01-07
 
 ### Added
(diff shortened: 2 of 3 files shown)


adamchainz:
@dlanderson care to review?

src/scout_apm/instruments/elasticsearch.py
```diff
@@ -84,11 +84,14 @@ def instrument_client(self):
 
         for method_str in self.__class__.CLIENT_METHODS:
             try:
-                code_str = """
+                code_str = """\
```
adamchainz: This is just to skip an unnecessary blank line in the temporary code, it makes line numbers a bit more understandable.

dlanderson: 👍

src/scout_apm/instruments/elasticsearch.py
```diff
@@ -84,11 +84,14 @@ def instrument_client(self):
...
 @monkeypatch_method(Elasticsearch)
 def {method_str}(original, self, *args, **kwargs):
     tr = TrackedRequest.instance()
-    index = kwargs.get('index', 'Unknown').title()
+    index = kwargs.get('index', 'Unknown')
+    if isinstance(index, (list, tuple)):
+        index = ','.join(index)
```
dlanderson: So with an array, we could end up with a name like `posts,articles` or something? This name is used as the layer name in our UI, so we probably want to figure out a better way to handle multiple strings from the array here.

adamchainz:
Good points about title casing. On further thought I think it could be removed, unless Scout really needs the values to be title-cased?

The `index` parameter may be a string containing a comma-separated list of indexes, so title-casing was already failing to title-case such a list. [ES docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html) state that index names are always lowercase. Also some users will be using the [wildcard syntax](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html), special value `_all`, or [date math](https://www.elastic.co/guide/en/elasticsearch/reference/current/date-math-index-names.html) which won't mean the same when title-cased.

dlanderson:
Decided to leave the naming as-is. LGTM!

PR #157, opened 2019-03-31
Fix code style for elasticsearch tests

adamchainz:
Missed in #156.


adamchainz:
@dlanderson I missed this in #156 because Travis is set not to build PR's, only branches, and my previous PR was made from my own fork. I presume that is for security because Travis has PyPI credentials?

PR #158, opened 2019-04-17
Fix 'ignore' functionality on Django < 1.10

adamchainz:
This second middleware class for old versions of Django was missed when adding the functionality in #144.

Tested with local Django app on Django 1.9 and Python 2.7. I looked into adding tests but there's not any machinery atm for determining if requests would be tracked.


cschneid:
Good catch!

Issue #120, opened 2018-08-27
Support Queue timing in Python agent


dlanderson:
Checking the request headers for officially supported frameworks like Django and Flask should be straight forward.

We should also have a generic API interface for recording Queue time in other cases.

dlanderson:
@cschneid Should we make this a layer as a child of the parent span?

cschneid:
I think that will work. The Ruby agent creates a span named: `"QueueTime/Request"`

dlanderson:
@adamchainz Can you look into implementing? You can create a new child span under the span we start in the Django middleware: https://github.com/scoutapp/scout_apm_python/blob/master/src/scout_apm/django/middleware.py#L25

Please also add for Flask, Pyramid, and Bottle.

Here's the headers we look for and how the Ruby agent does it: https://github.com/scoutapp/scout_apm_ruby/blob/master/lib/scout_apm/layer_converters/request_queue_time_converter.rb#L5

adamchainz:
Sure thing!

PR #167, opened 2019-04-25
Implement request queue timing in Django

adamchainz:
Per #120.

### DIFF
--- a/src/scout_apm/django/middleware.py
+++ b/src/scout_apm/django/middleware.py
@@ -4,11 +4,47 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import logging
 
 from scout_apm.api.context import Context
+from scout_apm.compat import datetime_to_timestamp
 from scout_apm.core.ignore import ignore_path
 from scout_apm.core.remote_ip import RemoteIp
 from scout_apm.core.tracked_request import TrackedRequest
+from scout_apm.core.util import convert_ambiguous_timestamp_to_ns
 
 logger = logging.getLogger(__name__)
+
+
+def track_request_queue_time(request, tracked_request):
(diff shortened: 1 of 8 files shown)


adamchainz:
This is just a start, need to look into properly testing it.

src/scout_apm/django/middleware.py
```diff
@@ -20,14 +22,34 @@ def __init__(self, get_response):
...
+        )
+        if not header_value:
+            return
+
+        raw_start = self._queue_time_re.sub('', header_value)
+        parsed_start = datetime.utcfromtimestamp(float(raw_start[0:10] + '.' + raw_start[10:13]))
+
+        tr.start_span(
+            operation="QueueTime/Request",
+            start_time=parsed_start,
```
cschneid: This may cause issues if this time is before the request start time. Unsure how that works in the core agent, needs testing.

adamchainz: You're right. I'll be testing on my demo app before adding unit tests so will let you know. Also the ruby agent has a guard against the start time being in the future which I need to add here.

adamchainz:
Split the `tracked_requests` stuff into #171, that needs fixing and merging first.

adamchainz:
Just spotted that according to https://docs.newrelic.com/docs/apm/applications-menu/features/configuring-request-queue-reporting , the header is always in microseconds. This needs tweaking.

dlanderson:
@adamchainz Does this still need tweaking?

adamchainz:
Yes it does, I'm also waiting for the agent to support the tag, see slack channel.

adamchainz:
I've merged although the agent doesn't support the tag yet in order to prevent merge conflicts :)

PR #170, opened 2019-05-01
Refactor Django configuration

adamchainz:
The existing tests did not reset Django to a clean state each time, since the scout_apm app was being repeatedly re-installed and uninstalled, but it didn't do anything to reset settings on uninstall. This is what lead to me implementing idempotent middleware addition in #168.

This PR refactors the way configuration is done. The test Django app is initialized only once, with Scout already installed. Scout then synchronizes the Django settings into ScoutConfig, and listens to Django's `setting_changed` signal to keep them in sync. Rather than rely on reset_all at the end of each test, I added an assertion that no settings have been left in place by any test - if they all use `override_settings` to change the settings, this will be the case.

This has let me write tests that cover all the cases with the middleware idempotency.

### DIFF
--- a/src/scout_apm/django/apps.py
+++ b/src/scout_apm/django/apps.py
@@ -5,6 +5,7 @@ import logging
 
 from django.apps import AppConfig
 from django.conf import settings
+from django.test.signals import setting_changed
 
 import scout_apm.core
 from scout_apm.core.config import ScoutConfig
@@ -19,7 +20,8 @@ class ScoutApmDjangoConfig(AppConfig):
     verbose_name = "Scout Apm (Django)"
 
     def ready(self):
-        self.copy_config()
+        self.update_scout_config_from_django_settings()
+        setting_changed.connect(self.on_setting_changed)
 
         # Finish installing the agent. If the agent isn't installed for any
         # reason, return without installing instruments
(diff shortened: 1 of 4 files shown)


src/scout_apm/django/apps.py
```diff
@@ -17,8 +19,8 @@ class ScoutApmDjangoConfig(AppConfig):
     verbose_name = "Scout Apm (Django)"
 
     def ready(self):
-        # Copy django configuration to scout_apm's config
-        ConfigAdapter.install()
+        self.update_scout_config_from_django_settings()
+        setting_changed.connect(self.on_setting_changed)
```
cschneid: Double checking - this settings watcher works on older versions of Django?

adamchainz: Yes the signal has been there since at least 1.8 which is what we support ( https://github.com/django/django/blob/stable/1.8.x/django/core/signals.py#L6 ) and tests will check it works too

adamchainz:
Will merge #171 first, this can then fix some of the testing problems seen there.

adamchainz:
Rebased but tests are still failing, need to dig into them a bit more.

adamchainz:
Okay fixed now. I made it so only one middleware type is tested on each Django version - it was too much trouble trying to reset *all* the state cleanly between runs and switch between the two middleware types. I think even the way middleware is mutated by `AppConfig.ready()` isn't a generally good idea, we should really have the user list the middleware in their stack manually and use a Django check to ensure it's there. This will avoid surprising interactions with other packages - outside of using Django's test utils, settings are meant to be immutable.

cschneid:
> we should really have the user list the middleware in their stack manually

The tradeoff is a less magical installation process for users. The elixir install has a lot more steps and it causes more support issues where the customer slightly misreads the different installation instructions between versions. It's just easier to miss a subtle difference during install. We can't avoid it with elixir but we can in this case.

adamchainz:
Fair enough. We could maybe document that manual install of the middleware is possible - at least the automatic injection is idempotent now.

PR #171, opened 2019-05-03
Test tracked requests and fix username tracking on Django < 1.10

adamchainz:
Split from #167.

The `tracked_requests` fixture intercepts all the requests that would go into `RequestBuffer` so assertions can be made on them.

Adding these assertions has revealed a number of bugs, only some of which I'm fixing here:
* Repeated spans - the middleware seems is re-installed repeatedly in tests, but it doesn't uninstall itself, so the instruments were installed on top of each other a lot. I've fixed this for now by making the installs idempotent, like the middleware, but long term will be making the django app only get installed once during tests.
* Username tracking wasn't implemented in the old Django middleware
* The test to disable monitoring doesn't really disable it because Scout is fully installed by the time the test runs. Some refactoring is needed here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Tested on Django 2.2
 - Added PyPI Trove classifiers for supported Django versions
+- Track usernames on Django < 1.10
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


adamchainz:
Tests are failing because of Django 1.11 differences around what's getting tracked, needs some more investigation I think

PR #183, opened 2019-06-20
Fix Flask to track missed automatic OPTIONS requests

adamchainz:
* Add failing test `test_hello_options` that shows how Monkey-patched `dispatch_request` missed Flask's `provide_automatic_options` behaviour
* Remove `process_request` and `process_response` hooks which have been empty since #99
* Stop entirely monkey patching `dispatch_request` ([Flask source](https://github.com/pallets/flask/blob/d4b688bd035b9704b3168b28fed39c8fcfe3b997/src/flask/app.py#L1872c)) and instead wrap it with a `CallableProxy` around `wrapped_dispatch_request`. This fixes the current divergence around `provide_automatic_options` and guards against potential future divergence.
* Remove the `try/except Exception` around putting in place `CallableProxy`, since we're no longer wrapping each view function but only the one method. This will fail at startup if it's not possible, and that doesn't seem likely anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 - Tested on Django 2.2
 - Added PyPI Trove classifiers for supported Django versions
 - Track usernames on Django < 1.10
+- Fix Flask integration to track some missed requests such as automatic
+  `OPTONS` responses.
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


adamchainz:
Sure thing

adamchainz:
Smoke test successful

PR #186, opened 2019-06-21
Refactor away GitRevision class

adamchainz:
It's such a simple amount of code it is better as a method on the one class that uses it. Also rewrote the tests to be cleaner, fully controlling `os.environ` and resetting it by using `mock.patch.dict`.


src/scout_apm/core/git_revision.py
```diff
@@ -1,22 +0,0 @@
...
-from __future__ import absolute_import, division, print_function, unicode_literals
-
-import os
-
-# Note that this is called from Config Defaults, which is looked up after the
-# configs from environment variables. If you set SCOUT_REVISION_SHA, it will be
-# picked up in that step of config, not here.
-
-# Similarly, this can be overriden by the code level configuration (from Django
-# configs or similar) by using the "revision_sha" key.
```
adamchainz: Considering these comments I think `SCOUT_REVISION_SHA` should actually be dropped from the lookup to avoid it being fetched in two places.

adamchainz:
Yes, it always did before this class thanks to `ScoutConfigEnv` being higher in the hierarchy.

PR #188, opened 2019-06-21
Add Falcon integration

### LINKED ISSUE #179: Add Falcon Instrumentation middleware
We have a Falcon example in the docs ( https://docs.scoutapm.com/#falcon-example ) but it would be good to have an official middleware ( https://falcon.readthedocs.io/en/stable/api/middleware.html ) for auto-instrumentation.

adamchainz:
Fixes #179. Add a middleware that implements basic tracking of Falcon resources, with operations named after the current resource name plus the method called on it. The method name is retrieved through stack inspection, which can fail and in that case we use the raw HTTP method name instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,22 @@ __A Scout account is required. [Signup for Scout](https://scoutapm.com/users/sig
 pip install scout-apm
 ```
 
+### Bottle
+
+```python
+from scout_apm.bottle import ScoutPlugin
+
+app = bottle.default_app()
+app.config.update({
+    "scout.name": "YOUR_APP_NAME",
+    "scout.key": "YOUR_KEY",
+    "scout.monitor": "true",
+})
+
+scout = ScoutPlugin()
+bottle.install(scout)
+```
+
 ### Django
 
 ```python
@@ -43,6 +59,22 @@ INSTALLED_APPS = [
 SCOUT_MONITOR = True
 SCOUT_KEY = "[AVAILABLE IN THE SCOUT UI]"
 SCOUT_NAME = "A FRIENDLY NAME FOR YOUR APP"
+```
+
+### Falcon
+
+```python
+import falcon
+from scout_apm.falcon import ScoutMiddleware
+
+scout_middleware = ScoutMiddleware(config={
+    "key": "[AVAILABLE IN THE SCOUT UI]",
+    "monitor": True,
+    "name": "A FRIENDLY NAME FOR YOUR APP",
+})
+api = falcon.API(middleware=[ScoutMiddleware()])
+# Required for accessing API internals, Falcon doesn't provide it default
+scout_middleware.set_api(api)
 ```
 
 ### Flask
@@ -85,22 +117,6 @@ if __name__ == "__main__":
         config.include("scout_apm.pyramid")
 
         # Rest of your config...
-```
-
-### Bottle
-
-```python
-from scout_apm.bottle import ScoutPlugin
-
-app = bottle.default_app()
-app.config.update({
-    "scout.name": "YOUR_APP_NAME",
-    "scout.key": "YOUR_KEY",
-    "scout.monitor": "true",
-})
-
-scout = ScoutPlugin()
-bottle.install(scout)
 ```
 
 For full installation instructions, including information on configuring Scout via environment variables, see our [Python docs](https://docs.scoutapm.com/#python-agent).
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ scout_middleware = ScoutMiddleware(config={
     "name": "A FRIENDLY NAME FOR YOUR APP",
 })
 api = falcon.API(middleware=[ScoutMiddleware()])
-# Required for accessing API internals, Falcon doesn't provide it default
+# Required for accessing extra per-request information
 scout_middleware.set_api(api)
 ```
 
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ Python Versions:
 Scout APM works with the following frameworks:
 
 * Django 1.8+
+* Falcon 2.0+
 * Flask 0.10+
 * Celery 3.1+
 * Pyramid 1.8+
--- a/src/scout_apm/falcon.py
+++ b/src/scout_apm/falcon.py
@@ -1,0 +1,87 @@
+# coding=utf-8
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import inspect
+
+from scout_apm.api import install
+from scout_apm.core.ignore import ignore_path
+from scout_apm.core.tracked_request import TrackedRequest
+
+# Falcon Middleware docs:
+# https://falcon.readthedocs.io/en/stable/api/middleware.html
+
+
+class ScoutMiddleware(object):
+    """
+    Falcon Middleware for integration with Scout APMi.
+    """
+
+    def __init__(self, config):
+        install(config=config)
+
+    def process_request(self, req, resp):
+        tracked_request = TrackedRequest.instance()
+        tracked_request.mark_real_request()
+        req.context.scout_tracked_request = tracked_request
+
+        tracked_request.tag("path", req.path)
+
+        if ignore_path(req.path):
+            tracked_request.tag("ignore_transaction", True)
+
+    def process_resource(self, req, resp, resource, params):
+        tracked_request = getattr(req.context, "scout_tracked_request", None)
+        if tracked_request is None:
+            # Somehow we didn't start a request - this might occur in
+            # combination with a pretty adversarial application, so guard
+            # against it, although if a request was started and the context was
+            # lost, other problems might occur.
+            return  # pragma: no cover
+
+        # Find the current responder's name. Falcon passes middleware the
+        # current resource but unfortunately not the method being called. Also
+        # middleware doesn't have a reference to the current API object so we
+        # can't check the router there. Use some hopefully pretty harmless
+        # stack inspection to look up the current responder method.
+        responder_name = None
+        current_frame = inspect.currentframe()
+        # current_frame can return None if stack inspection is not available
+        if current_frame is not None:
+            try:
+                if current_frame.f_back is not None:
+                    responder = current_frame.f_back.f_locals.get("responder", None)
+                    if responder is not None:
+                        responder_name = getattr(responder, "__name__", None)
+            finally:
+                # As per inspect docs, one should always be careful to release
+                # references to frames
+                del current_frame
+
+        if responder_name is not None:
+            operation = "Controller/{}.{}.{}".format(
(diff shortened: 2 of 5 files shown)


adamchainz:
Okay the approach with stack inspection has failed because Falcon is actually compiled with Cython. I didn't notice this locally because I didn't have Cython installed and the pure python wheel was installed on macOS. By installing Cython and then reinstalling from source I got the Cython compiled version and can reproduce the current failures locally as below. Investigating.

```
=================================== FAILURES ===================================
__________________________________ test_home ___________________________________
tracked_requests = [<TrackedRequest(req_id='req-498c6d23-c3f7-47a5-b327-aa52243c4653', tags={'path': '/', 'mem_delta': 0.0})>]
    def test_home(tracked_requests):
        with app_with_scout() as app:
            response = TestApp(app).get("/")
    
        assert response.status_int == 200
        assert response.text == "Welcome home."
        assert len(tracked_requests) == 1
        tracked_request = tracked_requests[0]
        assert tracked_request.tags["path"] == "/"
        assert tracked_request.active_spans == []
        assert len(tracked_request.complete_spans) == 1
        span = tracked_request.complete_spans[0]
>       assert (
            span.operation == "Controller/tests.integration.test_falcon.HomeResource.on_get"
        )
E       AssertionError: assert 'Controller/t...eResource/GET' == 'Controller/te...source.on_get'
E         Skipping 43 identical leading characters in diff, use -v to show
E         - meResource/GET
E         + meResource.on_get
tests/integration/test_falcon.py:70: AssertionError
______________________________ test_home_suffixed ______________________________
tracked_requests = [<TrackedRequest(req_id='req-04a149e8-bd55-4871-826b-05cd4dab5766', tags={'path': '/suffixed', 'mem_delta': 0.0})>]
    def test_home_suffixed(tracked_requests):
        with app_with_scout() as app:
            response = TestApp(app).get("/suffixed")
    
        assert response.status_int == 200
        assert response.text == "Welcome home, suffixed."
        assert len(tracked_requests) == 1
        tracked_request = tracked_requests[0]
        assert tracked_request.tags["path"] == "/suffixed"
        assert tracked_request.active_spans == []
        assert len(tracked_request.complete_spans) == 1
        span = tracked_request.complete_spans[0]
>       assert (
            span.operation
            == "Controller/tests.integration.test_falcon.HomeResource.on_get_suffixed"
        )
E       AssertionError: assert 'Controller/t...eResource/GET' == 'Controller/te..._get_suffixed'
E         Skipping 43 identical leading characters in diff, use -v to show
E         - meResource/GET
E         + meResource.on_get_suffixed
tests/integration/test_falcon.py:127: AssertionError
__________________________________ test_crash __________________________________
tracked_requests = [<TrackedRequest(req_id='req-b5d7106c-3d9e-4623-93ec-d7d539f7099c', tags={'path': '/crash', 'error': 'true', 'mem_delta': 0.0})>]
    def test_crash(tracked_requests):
        with app_with_scout() as app:
            response = TestApp(app).get("/crash", expect_errors=True)
    
        assert response.status_int == 748
        assert response.text == ""
        assert len(tracked_requests) == 1
        tracked_request = tracked_requests[0]
        assert tracked_request.tags["path"] == "/crash"
        assert tracked_request.tags["error"] == "true"
        assert tracked_request.active_spans == []
        assert len(tracked_request.complete_spans) == 1
        span = tracked_request.complete_spans[0]
>       assert (
            span.operation
            == "Controller/tests.integration.test_falcon.CrashResource.on_get"
        )
E       AssertionError: assert 'Controller/t...hResource/GET' == 'Controller/te...source.on_get'
E         Skipping 44 identical leading characters in diff, use -v to show
E         - shResource/GET
E         + shResource.on_get
tests/integration/test_falcon.py:214: AssertionError
```

adamchainz:
Okay I refactored it to have a slightly clunkier but more accurate setup process. The middleware now needs `set_api()` calling before it will allow requests through, so it has a reference back to the `falcon.API` object. I added README documentation for it.

README.md
```diff
@@ -44,6 +61,22 @@ SCOUT_KEY = "[AVAILABLE IN THE SCOUT UI]"
...
+import falcon
+from scout_apm.falcon import ScoutMiddleware
+
+scout_middleware = ScoutMiddleware(config={
+    "key": "[AVAILABLE IN THE SCOUT UI]",
+    "monitor": True,
+    "name": "A FRIENDLY NAME FOR YOUR APP",
+})
+api = falcon.API(middleware=[ScoutMiddleware()])
+# Required for accessing API internals, Falcon doesn't provide it default
```
dlanderson: "by default"

src/scout_apm/falcon.py
```diff
@@ -0,0 +1,90 @@
...
+except ImportError:  # pragma: no cover
+    falcon = None
+
+# Falcon Middleware docs:
+# https://falcon.readthedocs.io/en/stable/api/middleware.html
+
+
+class ScoutMiddleware(object):
+    """
+    Falcon Middleware for integration with Scout APMi.
```
dlanderson: "APMi" -> "APM"

src/scout_apm/falcon.py
```diff
@@ -0,0 +1,90 @@
...
+        self.api = None
+
+    def set_api(self, api):
+        if not isinstance(api, falcon.API):
+            raise ValueError("api should be an instance of falcon.API")
+        self.api = api
+
+    def process_request(self, req, resp):
+        if self.api is None:
+            raise RuntimeError(
```
dlanderson: I think historically we have logged failure without raising exceptions so that we never break an application in case of a configuration or unexpected error.

src/scout_apm/falcon.py
```diff
@@ -0,0 +1,90 @@
...
+                "Call {}.set_api() before requests begin".format(
+                    self.__class__.__name__
+                )
+            )
+
+        tracked_request = TrackedRequest.instance()
+        tracked_request.mark_real_request()
+        req.context.scout_tracked_request = tracked_request
```
dlanderson: Is there a possibility that `req.context` and `TrackedRequest.instance()` return a different TrackedRequest?

src/scout_apm/falcon.py
```diff
@@ -0,0 +1,90 @@
...
+        # security sensitive context.
+        user_ip = (
+            req.get_header("x-forwarded-for", default="").split(",")[0]
+            or req.get_header("client-ip", default="").split(",")[0]
+            or req.remote_addr
+        )
+        tracked_request.tag("user_ip", user_ip)
+
+    def process_resource(self, req, resp, resource, params):
+        tracked_request = getattr(req.context, "scout_tracked_request", None)
```
dlanderson: Same question as above re: TrackedRequest.instance() vs req.context

src/scout_apm/falcon.py
```diff
@@ -0,0 +1,90 @@
...
+        if tracked_request is None:
+            # Somehow we didn't start a request - this might occur in
+            # combination with a pretty adversarial application, so guard
+            # against it, although if a request was started and the context was
+            # lost, other problems might occur.
```
dlanderson: Anything that would raise exceptions?

src/scout_apm/falcon.py
```diff
@@ -0,0 +1,90 @@
...
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from scout_apm.api import install
+from scout_apm.core.ignore import ignore_path
+from scout_apm.core.tracked_request import TrackedRequest
+
+try:
+    import falcon
+except ImportError:  # pragma: no cover
+    falcon = None
```
dlanderson: Is this used?

src/scout_apm/falcon.py
```diff
@@ -0,0 +1,90 @@
...
+            # combination with a pretty adversarial application, so guard
+            # against it, although if a request was started and the context was
+            # lost, other problems might occur.
+            return
+
+        # Find the current responder's name. Falcon passes middleware the
+        # current resource but unfortunately not the method being called, hence
+        # we have to go through routing again.
+        responder, _params, _resource, _uri_template = self.api._get_responder(req)
+        operation = "Controller/{}.{}.{}".format(
```
dlanderson: What would an example of this look like? Are we going to have a high cardinality of `operation` names resulting from this?

README.md
adamchainz: Right, this doesn't make sense to say - it doesn't provide it all.

src/scout_apm/falcon.py
adamchainz: Thanks

src/scout_apm/falcon.py
adamchainz: I just did this so that the module would remain importable when falcon isn't installed, but I guess it's not needed since none of the other library integrations use this pattern.

src/scout_apm/falcon.py
adamchainz: I don't think so. You'll see in the tests e.g. `Controller/tests.integration.test_falcon.HomeResource.on_get` and `Controller/tests.integration.test_falcon.HomeResource.on_get_suffixed`. As I understand from the Falcon docs, suffixed methods are the normal way of writing multiple types of response for a single underlying resource, so I think it's best to include them in the operation name.

src/scout_apm/falcon.py
adamchainz: In adverse circumstances where some other code (i.e. a user integration with the API) finishes the request, yes. I think this way will die more reliably though in such adverse cases - the other integration that always goes through `TrackedRequest.instance()` can split data across separate TrackedRequests which might be intended for different URL's. With this approach, we could just see double tracking of some requests as they will `finish()` twice.

src/scout_apm/falcon.py
adamchainz: Not that I know of - we're just returning early if somehow `process_request` didn't seem to have been called. It doesn't look possible in a normal Falcon app - the routing logic is pretty clear: https://github.com/falconry/falcon/blob/master/falcon/api.py#L219 . 

`test_middleware_deleting_scout_tracked_request` tests the case of an adversarial set of middlewares around `process_resource` that somehow replace the context then set it back.

src/scout_apm/falcon.py
adamchainz: Right. I thought this case might be different since it's a missing step in configuration, but I guess it's not, so I'll move to logging an error if `self.api` is `None` and then having an alternative path for naming the operation from just the HTTP method which ignores the suffix and thus which Python method is called i.e. `Controller/tests.integration.test_falcon.HomeResource.on_get` would be just `Controller/tests.integration.test_falcon.HomeResource.GET`.

adamchainz:
Smoke test worked with new test app on https://github.com/adamchainz/scout-test-apps 
Screenshot 2019-06-25 at 16 10 55

PR #195, opened 2019-06-27
Improve Pyramid integration

adamchainz:
* Rearrange `try/except` so `stop_span()` will only be called if it was started
* Use and test the full `user_ip` algorithm rather than just `request.remote_addr`, and reduce scope of its `try/except`
* Improve some tests

### DIFF
--- a/src/scout_apm/pyramid.py
+++ b/src/scout_apm/pyramid.py
@@ -2,7 +2,6 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import scout_apm.core
-from scout_apm.api.context import Context
 from scout_apm.core.config import ScoutConfig
 from scout_apm.core.ignore import ignore_path
 from scout_apm.core.tracked_request import TrackedRequest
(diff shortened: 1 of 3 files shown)


src/scout_apm/pyramid.py
```diff
@@ -24,34 +23,42 @@ def includeme(config):
...
-            span = tr.start_span(operation="Controller/Pyramid")
+        tracked_request = TrackedRequest.instance()
+        span = tracked_request.start_span(operation="Controller/Pyramid")
 
+        try:
             if ignore_path(request.path):
-                tr.tag("ignore_transaction", True)
+                tracked_request.tag("ignore_transaction", True)
+
+            tracked_request.tag("path", request.path)
```
dlanderson: Any reason we moved this outside the `try`? `.path` should supposedly always be present, but it's not critical in order for the TrackedRequest to work, and there's no harm in having it within the `try` where we do `user_ip` as well, right?

adamchainz: `request.path` was already accessed outside of the `try` above - when I started writing a test to make it crash, I realized this.

PR #202, opened 2019-07-01
Fix core-agent-manager

### LINKED ISSUE #120: Support Queue timing in Python agent

adamchainz:
Fixes #120. Steps:

* Added a call to `AgentContext.build()` to restore functionality. This is the
  only thing that broke. I used `git bisect` to discover that it was broken in
  6ab579e22ab903bb6b72108bc485550bb151f353, which is when `AgentContext` was
  refactored.
* Refactored to remove the unnecessary extra functions for the subparsers
* Removed unnecessary subparser name validity check, as this is done by
  argparse already
* Restored test coverage checking for the file
* Added some basic tests to get to 100% coverage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Fixed
 
 - Track path and user IP on Django < 1.10
+- The undocumented `core-agent-manager` CLI command works again
 
 ## [2.1.0] 2019-06-25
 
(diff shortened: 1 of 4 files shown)


adamchainz:
Do you think it's worth documenting now then?

PR #205, opened 2019-07-02
Add extra Celery tags

### LINKED ISSUE #34: Capture Celery Queue Name

adamchainz:
Fixes #34. Tagging with the current *queue* is only *sometimes* possible - often the *exchange* and *routing_key* are provided instead. The task may have come from a particular queue, but its delivery information normally onyl tracks where it was sent at the top of the AMQP funnel by exchange + routing_key, since that's what's needed for retry. Also track the `is_eager` flag which may be useful for determining which way the task was run.

More information on these parameters can be gleaned from [the source](https://github.com/celery/celery/blob/8cf8fae70630954cff3448485588bbe2a77ff3ab/celery/app/task.py#L600).

Updating the tests to use the fixture added in Celery 4.0, but since we support 3.1 this is in a conditional test. Our test suite currently installs the latest version of Celery only, so I tested locally temporarily with Celery 3.1 as well by specifying `celery>=3.1,<3.2` in `tox.ini`. For complete automated testing in the future, we should look at a better requirements management system.

### DIFF
--- a/src/scout_apm/celery.py
+++ b/src/scout_apm/celery.py
@@ -7,20 +7,22 @@ import scout_apm.core
 from scout_apm.core.tracked_request import TrackedRequest
 
 
-# TODO: Capture queue.
-# https://stackoverflow.com/questions/22385297/how-to-get-the-queue-in-which-a-task-was-run-celery?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa
-def prerun_callback(sender=None, headers=None, body=None, **kwargs):
-    name = kwargs["task"].name
+def prerun_callback(task=None, **kwargs):
+    tracked_request = TrackedRequest.instance()
+    tracked_request.mark_real_request()
 
-    tr = TrackedRequest.instance()
-    tr.mark_real_request()
-    span = tr.start_span(operation=("Job/" + name))
-    span.tag("queue", "default")
+    delivery_info = task.request.delivery_info
+    tracked_request.tag("is_eager", delivery_info.get("is_eager", False))
+    tracked_request.tag("exchange", delivery_info.get("exchange", "unknown"))
+    tracked_request.tag("routing_key", delivery_info.get("routing_key", "unknown"))
+    tracked_request.tag("queue", delivery_info.get("queue", "unknown"))
+
+    tracked_request.start_span(operation=("Job/" + task.name))
 
 
-def postrun_callback(sender=None, headers=None, body=None, **kwargs):
-    tr = TrackedRequest.instance()
-    tr.stop_span()
+def postrun_callback(task=None, **kwargs):
+    tracked_request = TrackedRequest.instance()
+    tracked_request.stop_span()
 
 
 def install():
--- a/tests/integration/test_celery.py
+++ b/tests/integration/test_celery.py
@@ -3,25 +3,41 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 from contextlib import contextmanager
 
-from celery import Celery
+import celery
+import pytest
+from celery.signals import setup_logging
 
 import scout_apm.celery
 from scout_apm.api import Config
 
+# http://docs.celeryproject.org/en/latest/userguide/testing.html#py-test
+skip_unless_celery_4_plus = pytest.mark.skipif(
+    celery.VERSION < (4, 0), reason="pytest fixtures added in Celery 4.0"
+)
+
+
+@setup_logging.connect
+def do_nothing(**kwargs):
+    # Just by connecting to this signal, we prevent Celery from setting up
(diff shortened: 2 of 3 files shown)


tests/integration/test_celery.py
```diff
@@ -3,27 +3,42 @@
...
+def do_nothing(**kwargs):
+    # Just by connecting to this signal, we prevent Celery from setting up
+    # logging - and stop it from interfering with global state
+    # http://docs.celeryproject.org/en/v4.3.0/userguide/signals.html#setup-logging
+    pass
+
 
 @contextmanager
-def app_with_scout(config=None):
+def app_with_scout(app=None, config=None):
```
dlanderson: How does changing this method signature affect current calls to this function? Who calls this and from where?

src/scout_apm/celery.py
```diff
@@ -7,20 +7,21 @@
 from scout_apm.core.tracked_request import TrackedRequest
 
 
-# TODO: Capture queue.
-# https://stackoverflow.com/questions/22385297/how-to-get-the-queue-in-which-a-task-was-run-celery?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa
-def prerun_callback(sender=None, headers=None, body=None, **kwargs):
-    name = kwargs["task"].name
+def prerun_callback(task=None, **kwargs):
```
dlanderson: Changed method signature. Who calls this and from where?

src/scout_apm/celery.py
```diff
@@ -7,20 +7,21 @@
...
+    span.tag("is_eager", delivery_info.get("is_eager", False))
+    span.tag("exchange", delivery_info.get("exchange", "unknown"))
+    span.tag("routing_key", delivery_info.get("routing_key", "unknown"))
+    span.tag("queue", delivery_info.get("queue", "unknown"))
 
 
-def postrun_callback(sender=None, headers=None, body=None, **kwargs):
-    tr = TrackedRequest.instance()
-    tr.stop_span()
+def postrun_callback(task=None, **kwargs):
```
dlanderson: Changed method signature. Who calls this and from where?

src/scout_apm/celery.py
```diff
@@ -7,20 +7,21 @@
...
-    tr = TrackedRequest.instance()
-    tr.mark_real_request()
-    span = tr.start_span(operation=("Job/" + name))
-    span.tag("queue", "default")
+    span = tracked_request.start_span(operation=("Job/" + task.name))
+    delivery_info = task.request.delivery_info
+    span.tag("is_eager", delivery_info.get("is_eager", False))
+    span.tag("exchange", delivery_info.get("exchange", "unknown"))
+    span.tag("routing_key", delivery_info.get("routing_key", "unknown"))
+    span.tag("queue", delivery_info.get("queue", "unknown"))
```
dlanderson: I think the core agent knows how to handle the `queue` tag, but the others are not supported explicitly and will be silently dropped, IIRC.  Are we wanting to track this info in span tags or in context? @cschneid might chime in here.

src/scout_apm/celery.py
adamchainz: This is called by Celery itself, through its [signals](http://docs.celeryproject.org/en/v4.3.0/userguide/signals.html), as set up below in `install()`. It's always called with kwargs, so changing the signature here doesn't do anything for the caller, just inside the body of the function we only bind `task` and ignore the other parameters. According to [the docs for `task_prerun`](http://docs.celeryproject.org/en/v4.3.0/userguide/signals.html#task-prerun) the `headers` and `body` kwargs are not passed anyway, and `sender` is equal to `task`

src/scout_apm/celery.py
adamchainz: Aha, I didn't know that span tags are dropped. Moving to request tags/context is probably the right move in that case, but I'll wait for @cschneid 's input.

src/scout_apm/celery.py
adamchainz: It's in the same position as prerun_callback above

src/scout_apm/celery.py
cschneid: Currently in the core agent, span tags are very specific, and unknown ones are dropped. Some upcoming work will be to support the new trace structure, where arbitrary span tags are sent onward, but that's not for a bit. `queue` is recognized as the string queue name that the job was enqueued on & then popped off of.

For now it's probably ok to add these as request tags (context). But anyway, this does feel like it's scoped to the whole request (the job), not just the section of code that celery is calling first..

adamchainz: Moving them all to request tags then. I think I had a concern that celery might recursively execute tasks with its fancy chord structures, but after checking the docs again I don't think that's the case

PR #206, opened 2019-07-02
Track queue time on Celery

### LINKED ISSUE #35: Capture Celery Time-in-queue

adamchainz:
Fixes #35. Track the queue time by using an extra haeder to track when the task is added to the queue and comparing when it gets executed. This should work in most setups but if the celery app isn't all set up on the producer side, we are well defended against it.

Like #205, I manually tested on Celery 3.1.x locally by editing `tox.ini` and reinstalling with `tox -r`.

### DIFF
--- a/src/scout_apm/celery.py
+++ b/src/scout_apm/celery.py
@@ -1,15 +1,33 @@
 # coding=utf-8
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from celery.signals import task_postrun, task_prerun
+import datetime as dt
+
+from celery.signals import before_task_publish, task_postrun, task_prerun
 
 import scout_apm.core
+from scout_apm.compat import datetime_to_timestamp
 from scout_apm.core.tracked_request import TrackedRequest
+
+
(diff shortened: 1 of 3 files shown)


src/scout_apm/celery.py
```diff
@@ -1,17 +1,36 @@
...
+
+    start = getattr(task.request, "scout_task_start", None)
+    if start is not None:
+        now = datetime_to_timestamp(dt.datetime.utcnow())
+        try:
+            queue_time = now - start
+        except TypeError:
+            pass
+        else:
+            span.tag("queue_time", queue_time)
```
adamchainz: I'm guessing this should be a request tag (aka context) instead of a span tag @dlanderson / @cschneid ? same as discussion on the other celery PR https://github.com/scoutapp/scout_apm_python/pull/205/files#r300129980

cschneid: Yes, a request tag is better here. Especially since it's not describing "the section of code that is running", it's saying something about the request as a whole.

PR #218, opened 2019-08-02
Use Django 2.0+ database instrumentation

### LINKED ISSUE #217: Use Django 2.0+ DB instrumentation hook
Rather than monkey-patching, use the built-in instrumentation on Django 2.0+:

https://docs.djangoproject.com/en/2.0/topics/db/instrumentation/

adamchainz:
Fixes #217.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Pending
+
+### Added
+
+- Use Django's native DB instrumentation on Django 2.0+, rather than monkey
+  patching the database cursor.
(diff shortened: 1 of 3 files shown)


adamchainz:
WIP right now - since execute wrappers are per-connection per-thread, there isn't an easy way I've found to install one globally. Still need to think this through.

adamchainz:
I updated to use a signal handler to install whenever a connection is made, and to install idempotently. This is bypassing the Django public API a bit, but I believe it's better than monkey patching.

PR #221, opened 2019-08-04
Fix path tagging on Flask

adamchainz:
This was previously being tagged on the span, rather than the tracked request itself, which meant it was not seen by the core agent / UI.


adamchainz:
Forgot to attach these..

before:
Screenshot 2019-08-04 at 11 14 35

after (there's a forward slash!):
Screenshot 2019-08-04 at 11 17 48

PR #223, opened 2019-08-07
Add Dramatiq integration

### LINKED ISSUE #222: Add support for Dramatiq
Could we investigate the feasibility of adding support for the Dramatiq background job library?

https://dramatiq.io/

We have only had the one request for it as far as I am aware. But if is not a huge job then it would be beneficial to have going forward.

adamchainz:
Fixes #222.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Pending
+
+### Added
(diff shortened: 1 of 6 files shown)


adamchainz:
Needs disabling on Python <3.5 it seems!

PR #242, opened 2019-08-29
Add test coverage for samplers

adamchainz:
This revealed a few bugs in `Cpu`:

* The first defensive log message was broken in terms of expected format parameters. I dropped the last one.
* The last branch defending against `res < 0` was unreacahable since the two divisors had already been verified as not being negative, so the division could not end up negative.

### DIFF
--- a/src/scout_apm/core/samplers/cpu.py
+++ b/src/scout_apm/core/samplers/cpu.py
@@ -33,7 +33,7 @@ class Cpu(object):
         if wall_clock_elapsed < timedelta(0):
             self.save_times(now, cpu_times)
             logger.debug(
-                "%s: Negative time elapsed. now: %s, last_run: %s, total time: %s.",
+                "%s: Negative time elapsed. now: %s, last_run: %s.",
                 self.human_name(),
                 now,
                 self.last_run,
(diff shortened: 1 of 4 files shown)


adamchainz:
Seeing some flaky test failures:

```
        assert result is None
>       assert len(caplog.record_tuples) == 1
E       AssertionError: assert 3 == 1
E        +  where 3 = len([('scout_apm.core.socket', 10, 'CoreAgentSocket attempt 2, connecting to /tmp/scout_apm_core/scout_apm_core-v1.1.11-x8...pu', 10, 'Process CPU: Negative time elapsed. now: 2019-08-29 08:11:56.343240, last_run: 2019-08-29 08:11:56.343240.')])
E        +    where [('scout_apm.core.socket', 10, 'CoreAgentSocket attempt 2, connecting to /tmp/scout_apm_core/scout_apm_core-v1.1.11-x8...pu', 10, 'Process CPU: Negative time elapsed. now: 2019-08-29 08:11:56.343240, last_run: 2019-08-29 08:11:56.343240.')] = <_pytest.logging.LogCaptureFixture object at 0x7f0714762ba8>.record_tuples

```

It looks like the core agent socket thread is left running in the background by other tests and this occasionally happens to log during these new tests.

I'm going to workaround for now by filtering the log messages in these tests, but this needs more work to make robust.

PR #243, opened 2019-08-29
Fix race condition in CoreAgentSocket.stop()

adamchainz:
I realized that this method has a read-after-write concurrency bug here - if the thread stops successfully but another restarts the thread between the call to `wait()` and `is_set()`, it could incorrectly report failure to stop. `wait()` already returns the value of the event so there's no need to reread here.


dlanderson:
👍 Looks good.

Issue #248, opened 2019-08-30
Allow configuration of 'hostname'


cschneid:
A customer has a setup where all of their servers have the same value for their hostnames. This causes the nodes table in our UI to merge all of their info.

* add a configuration value
* send that to Core Agent at startup.

Then there is a Core Agent task to pass that data along correctly.

adamchainz:
This is already supported. We have the [`hostname` key in config](https://github.com/scoutapp/scout_apm_python/blob/41cf5f682f757df868732601dc700891ff3e4413/src/scout_apm/core/config.py#L66). It's passed to the agent in [AppMetadata](https://github.com/scoutapp/scout_apm_python/blob/41cf5f682f757df868732601dc700891ff3e4413/src/scout_apm/core/metadata.py#L37).

I just tested my Django test app with `SCOUT_HOSTNAME=foobar python app.py runserver` - I could override it and it was passed to the core agent, but it didn't get passed through to Scout as my node name still showed up as my computer name.

I think this only needs core agent changes then?

adamchainz:
Fixed in #251.

nMustaki:
Thanks !

PR #251, opened 2019-09-11
Override hostname

naohiro-t:
See: https://github.com/scoutapp/scout_apm_python/issues/248
This also requires the binary agent changes, so do not merge yet.

### DIFF
--- a/src/scout_apm/core/commands.py
+++ b/src/scout_apm/core/commands.py
@@ -16,13 +16,15 @@ class Register(object):
     def __init__(self, *args, **kwargs):
         self.app = kwargs.get("app")
         self.key = kwargs.get("key")
+        self.hostname = kwargs.get("hostname")
 
     def message(self):
-        logger.info("Registering with app=%s key=%s" % (self.app, self.key))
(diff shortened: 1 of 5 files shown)


src/scout_apm/core/commands.py
```diff
@@ -16,13 +16,18 @@ class Register(object):
...
-        logger.info("Registering with app=%s key=%s" % (self.app, self.key))
+        logger.info(
+            "Registering with app=%s key=%s host=%s"
+            % (self.app, self.key, self.hostname)
+        )
         return {
             "Register": {
                 "app": self.app,
                 "key": self.key,
+                "host": self.hostname,
```
cschneid: The hostname arg needs to be optional. We should have `None` if it's not set

naohiro-t: Should we set default value to `None`?

adamchainz: Do you want to pass `host` or `hostname` to the core agent?

adamchainz: @cschneid verified on Slack that the agent expects `host`

PR #285, opened 2019-10-23
[WIP] Add ASGI integration

adamchainz:
Basics of an ASGI integration that will work cross-framework. Include some source that won't be included on old Python versions, tests that won't be collected on old Python versions, and test using `asgiref.testing.ApplicationCommunicator`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,10 @@ from setuptools import Extension, find_packages, setup
 with open("README.md", "r") as fp:
     long_description = fp.read()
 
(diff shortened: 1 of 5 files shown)


adamchainz:
I don't believe a generic ASGI integration is easy/possible right now. Whilst ASGI is very symmetric and "turtles all the way down," without framework specific stuff such as Starlette's `context["endpoint"]`, we can't find a good transaction name. Therefore I'm closing this right now, and we'll instead pursue one integration per framework.

PR #302, opened 2019-10-25
[WIP] Generic WSGI integration

adamchainz:


adamchainz:
So just like the generic ASGI integration I started in #285 I think this isn't possible to do very well. The main problem is finding good names for transactions. I think we're better off adding more framework support - the `WebTransaction` class in the API is 'good enough' for users tracking unsupported frameworks.

Issue #28, opened 2018-04-19
Gunicorn and gevent worker support


itsderek23:
We haven't tested the agent with [async workers via gevent](http://docs.gunicorn.org/en/stable/design.html#async-workers) yet. 

We should.

svvitale:
So my use case is slightly different with gunicorn + uvicorn (which is an async worker using uvloop under the hood).  I'm guessing that the issue may be with Async workers in general, so I thought I would post these details here.  If this belongs in its own issue, please let me know.

The setup for scout-apm appears to work correctly, and I see debug messages that look reasonable:

[scout_apm_debug.log](https://github.com/scoutapp/scout_apm_python/files/3267875/scout_apm_debug.log)

I never see any request data show up in the Scout dashboard though.

Python 3.6
Django 2.0.3
Channels 2.1.5
gunicorn 19.9.0
uvicorn 0.4.6

This is all running on Heroku and uses the following Procfile line for starting the web process:

`gunicorn server.asgi -w $WEB_WORKERS -k uvicorn.workers.UvicornWorker --log-file -`

asgi.py:
```
"""
ASGI entrypoint. Configures Django and then runs the application
defined in the ASGI_APPLICATION setting.
"""
import os
import django
from channels.routing import get_default_application

os.environ.setdefault("DJANGO_SETTINGS_MODULE", "server.settings")
django.setup()
application = get_default_application()
```

routing.py:
```
from django.conf.urls import url
from channels.routing import URLRouter
from channels.http import AsgiHandler
from channels.auth import AuthMiddlewareStack
import django_eventstream
from channels.routing import ProtocolTypeRouter

urlpatterns = [
    url(r"^channels/", AuthMiddlewareStack(URLRouter(django_eventstream.routing.urlpatterns))),
    url(r"", AsgiHandler),
]

application = ProtocolTypeRouter({"http": URLRouter(urlpatterns)})
```

Let me know if I can provide any further detail.  I've dug in as far as to verify that the middleware is being executed, but I'm not sure where else to look from here.

Thanks!

adamchainz:
Thanks for the detailed report @svvitale . I'll chat with the other scout members and prioritize supporting ASGI.

svvitale:
Just wanted to check-in and see if there's been any priority allocated to ASGI support?  I'm happy to be the guinea pig here.

adamchainz:
Hi @svvitale 

[...]

PR #349, opened 2019-11-04
Add Starlette integration

### LINKED ISSUE #233: Support for Starlette framework
Had a customer come through who was trying to use Scout on their app: 
Image 2019-08-23 at 3 19 28 PM

Customer said that the app is only reporting the 4XX and 5XX responses, not the 2XX

adamchainz:
Fixes #233 .

Basic Starlette integration that tracks HTTP only. Websocket implementation would be a bit more fun.

### DIFF
--- a/src/scout_apm/core/context.py
+++ b/src/scout_apm/core/context.py
@@ -1,14 +1,17 @@
 # coding=utf-8
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import threading
-
 from scout_apm.core.tracked_request import TrackedRequest
 
+try:
+    from asgiref.local import Local
+except ImportError:
+    from threading import local as Local
 
-class ThreadLocalContext(object):
+
+class LocalContext(object):
     def __init__(self):
-        self._context = threading.local()
+        self._context = Local()
 
     def get_tracked_request(self):
         if not hasattr(self._context, "tracked_request"):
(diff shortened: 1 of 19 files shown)


src/scout_apm/core/context.py
```diff
@@ -1,14 +1,17 @@
...
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import threading
-
 from scout_apm.core.tracked_request import TrackedRequest
 
+try:
+    from asgiref.local import Local
+except ImportError:
+    from threading import local as Local
```
dlanderson: So if they are not running Python 3.5+ with asgiref, we fall back to using Treading.local, which will not support async tracking, correct?

I wonder if we should log a line that syas something to that effect: `Async tracking unavailable (Requires Python 3.5+)` or similar

adamchainz: It's only in Python 3.5 that the async functionality was added to Python so that wouldn't really add any information :)

dlanderson: Ah, I had thought Python had async for longer. Nevermind, then!

adamchainz: There's been gevent and other styles of async for a while - in general it's a confusing landscape. Still need to get around to official gevent support in #28.

PR #375, opened 2019-11-12
Instrument RQ

### LINKED ISSUE #370: Instrument rq
RQ is a popular task queue: https://python-rq.org/

We should have an integration just like we do other task queues.

adamchainz:
Fixes #370.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Pending
+
+### Added
+
+- Add RQ integration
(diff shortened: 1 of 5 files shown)


adamchainz:
Want to write the docs before I merge this. Had to consider a few designs, eventually settled on a custom `Worker` class with decorator-patching of `Job` that seems to do everything we'd want.

adamchainz:
Docs are up at https://github.com/scoutapp/slate_apm_help/pull/70 . I've retested my [rq test apps](https://github.com/adamchainz/scout-test-apps) and am happy with this now!

PR #380, opened 2019-11-15
Improve Django Rest Framework ViewSet transaction names

### LINKED ISSUE #102: Instrument the Django REST Framework (DRF)
Provide more detailed instrumentation of the Django REST Framework (DRF).

http://www.django-rest-framework.org/

adamchainz:
Fixes #102.


ryanhiebert:
Thanks for doing this, @adamchainz . I think this is going to be really cool. I saw it in the release log today, so looks like it'll be getting on my production servers shortly.

adamchainz:
👍 Let me know how it goes!

Issue #383, opened 2019-11-21
Track all 5XX responses as errors


adamchainz:
Currently our web integrations we tend to insert ourselves as a "middleware" layer, and if an exception is raised below us, we catch it, track the current request as an error, then re-raise it. This allows the web framework to turn the exception into status code 5XX response.

Sometimes in web frameworks, a layer can be added below us that already catches exceptions and returns status code 5XX responses. For example, Django REST Framework does this - see #102.

We currently count such requests as successful. Instead we should tag them as errors.

adamchainz:
Decided it's not just 500 but all 5XX - sometimes web apps return e.g. 503 because they can't get through to a data store.

adamchainz:
All web frameworks now done!

PR #390, opened 2019-11-23
Track Flask 5XX responses as errors

adamchainz:
Ref #383.


adamchainz:
WIP - there are many cases to handle from Flask, or we move to wrapping at a different function.

adamchainz:
Moved to wrapping one level up in `full_dispatch_request`. This has the side effect of tracking responses generated by preprocessors.

PR #395, opened 2019-11-25
Track Flask preprocess_request span

adamchainz:
Following #390 I realized it's possible to track this middleware-ish hook which might be of value to our Flask users.


adamchainz:
It might be better to move to tracking the preprocessors as separate span around the actual request processing.

PR #403, opened 2019-11-27
Remove passenv = * from tox config

adamchainz:
Only pass the environment variables we know the test suite uses. This prevents unintended overrides. `SCOUT_*` env vars are already cleared in `conftest.py` thanks to 475620d3 but this is a more solid approach.


adamchainz:
Test failure is the one I'm fixing in #404

PR #412, opened 2019-12-02
[WIP] Instrument Django Channels

### LINKED ISSUE #372: Instrument Django Channels
This issue mostly discussed some bugs relating to async channels behaviour: #46 . It also noted we could instrument Channels better.

We already instrument normal Django views run under Channels because the full Django middleware stack runs. We'd better just look at instrumenting other types of consumers - notably websockets and async http consumers.

adamchainz:
Fixes #372.


adamchainz:
WIP - currently only instrumenting Async HTTP Consumer, and very incompletely. Will want to instrument the methods inside the generic WebSocket consumer too.

adamchainz:
My current approach of decorating all the built-in consumer classes is ending up with lots of duplication. I went this way because it looked like there was no easy way of building a middleware. However the main reason for this is because Channels is on ASGI 2 and if it was upgraded to ASGI 3 (issue: https://github.com/django/channels/issues/1316 ) we could much more easily provide a single unified middleware.

That said, there's little work being done on this, and many customers are likely to remain on older Channels versions since the change might be disruptive to their apps.

It might be possible to build a middleware class that works for both ASGI 2 and 3 so it would continue to work going forwards.

adamchainz:
Closing for now as we don't have a good way of working with websockets.

PR #414, opened 2019-12-03
Use Django's apps.is_installed in Huey instrumentation

adamchainz:
This is the "correct" way to do it.


adamchainz:
This doesn't work because we're trying to check at the time that Huey is being configured. Potentially could be considered a bug in Django.. but meh, it works checking `INSTALLED_APPS` too.

PR #450, opened 2020-01-03
Rewrite CoreAgentSocket as a SingletonThread

adamchainz:
Refactor in order to:

* Use the simpler, tested, idempotent thread logic of `SingletonThread`, extracted from `SamplersThread` in #448.
* Use a shared command queue - commands enqueued while the thread happens to be shutting down will still be sent.
* Provide a simpler interface to users - simply call `send(command)`.
* Avoid `ResourceWarning`s from sockets being left open in tests - they should all be `close()`d now.

### DIFF
--- a/src/scout_apm/core/__init__.py
+++ b/src/scout_apm/core/__init__.py
@@ -10,7 +10,7 @@ from scout_apm.core import objtrace
 from scout_apm.core.config import scout_config
 from scout_apm.core.core_agent_manager import CoreAgentManager
 from scout_apm.core.metadata import report_app_metadata
-from scout_apm.core.socket import CoreAgentSocket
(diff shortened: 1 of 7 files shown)


adamchainz:
For the record, here's the `ResourceWarning`s I was seeing previously:

```

tests/integration/core/test_socket.py::test_socket_instance_is_recreated_if_not_running

  /home/travis/build/scoutapp/scout_apm_python/.tox/py36-django111/lib/python3.6/site-packages/scout_apm/core/socket.py:35: ResourceWarning: unclosed <socket.socket fd=24, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0>

    cls._instance = cls()
```

PR #452, opened 2020-01-21
Ignore false positive from flake8-bugbear B014

adamchainz:
As per referenced issue, OSError and IOError are only the same thing on Python 3, but we support Python 2.


adamchainz:
Test failures unrelated.

Issue #336, opened 2019-11-03
Instrument hug


adamchainz:
Hug was reported used by 1% of developers in the [Python developers survey 2018](https://www.jetbrains.com/research/python-developers-survey-2018/). We should support it.

PR #453, opened 2020-01-21
Improve Falcon integration

adamchainz:
Refs #336. Hug uses Falcon for its HTTP API interface. This PR contains some improvements to the Falcon integration that will be helpful for Hug:

* `set_api()` is no longer required, thanks to some stack inspection.
* Tracing is automatically disabled if installation fails, like other intgerations.
* Callable class-based responders are supported. Hug uses one of these.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,8 @@
 
 ### Added
 
-- Improved Falcon integration: `set_api()`  is no longer required, tracing
-  is automatically disabled if installation fails, and callable class-based
-  responders are supported.
+- Improved Falcon integration: tracing is automatically disabled if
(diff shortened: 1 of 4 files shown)


adamchainz:
Okay so after some debugging - `set_api()` removal doesn't work on CI because there we're using the Cythonized version, for which stack frames don't exist. It seems there's no reliable way to automatically discover the current Falcon API given that stack frames don't exist. This is a [documented Cython limitation](http://docs.cython.org/en/latest/src/userguide/limitations.html). Reverting so that users will still need to call `set_api()`.

adamchainz:
The only test failures are Jinja2 async that are also happening on master.

PR #455, opened 2020-01-21
Workaround Elasticsearch scripts_painless_context disappearance

adamchainz:


adamchainz:
Test failure from another Elasticsearch failure.

Issue #456, opened 2020-01-21
Elasticsearch 7.5.1 changed positional arguments order


adamchainz:
Similar to #454, Elasticsearch 7.5.1 broke the test suite in another way.

https://github.com/elastic/elasticsearch-py/issues/1099

PR #457, opened 2020-01-21
xfail Elasticsearch positional argument test

adamchainz:
Refs #456.


adamchainz:
Test failures here should be fixed by #458.

PR #462, opened 2020-01-29
Fix Jinja2 async integration for Jinja2 2.11.0+

### LINKED ISSUE #461: Fix Jinja2 async integration
Since https://github.com/pallets/jinja/issues/765 , Jinja2 only lazily imports and adds the `render_async` function that we instrument (since #398).

We need to update to only instrument at the point the `render_async` method becomes available.

adamchainz:
Fixes #461.


adamchainz:
Test error is the redis one fixed in #463

PR #474, opened 2020-02-14
Flush transactions at shutdown

### LINKED ISSUE #473: Add a "flush background thread" function
To support one-off cron tasks that exit python when finished, we need to allow the background thread to finish its work. Currently, if the main thread exits, any in-flight data the background thread has is lost.

We should add a `flush_sync` or similar function that blocks until the background thread has sent whatever payloads it has over to the core agent.

adamchainz:
Fixes #473.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@
   ([PR #460](https://github.com/scoutapp/scout_apm_python/pull/460)).
 - Limit size of recorded tracebacks to reduce memory usage
   ([PR #467](https://github.com/scoutapp/scout_apm_python/pull/467)).
+- Scout now flushes any queued commands to the core agent at shutdown, with a
+  configurable timeout `shutdown_timeout_seconds`, defaulting to 2 seconds.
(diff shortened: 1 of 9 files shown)


adamchainz:
So this is half-done right now

I think having it on by default using Python's built-in main thread `atexit.register` would be good, with a stderr message. Prior art here is Sentry: https://github.com/getsentry/sentry-python/blob/f763061ed9d9e99d85b3e95adc3ed63b623fc4a0/sentry_sdk/integrations/atexit.py#L30 . It has a short timeout and I don't think it's ever been annoying on any of the 10+ projects I've used it on. We'd add a new configuration variable fo seconds to wait at shutdown, which if 0 would mean no wait. Then users with cron jobs wouldn't need to make any change.

What do you think @cschneid ?

adamchainz:
I've refactored to automatically install the `atexit` handler, but we could easily change that to be opt-out by setting the default `shutdown_timeout_seconds` to 0.0. I think copying Sentry with a default 2 second timeout should be fine myself though.

PR #476, opened 2020-02-14
Limit traceback size

adamchainz:
Copied from the Ruby agent: https://github.com/scoutapp/scout_apm_ruby/blob/84575a57f2fb64bced6f2394632212f2bfb0268f/lib/scout_apm/layer.rb#L116

Not strictly related to implementing the issue, but spotted whilst investigating it as an idea to reduce memory usage and overhead.


adamchainz:
(Python's default stack size is 1000 which is probably TMI)

PR #479, opened 2020-02-17
Move to PyPI API token

adamchainz:
As per https://pyfound.blogspot.com/2020/01/start-using-2fa-and-api-tokens-on-pypi.html .


adamchainz:
Travis won't discover any problem with this until release time...

PR #480, opened 2020-02-17
Move from requests to urllib3

adamchainz:
Basically avoid using `requests` so that we aren't pulling it in for people who aren't using it. Whilst popular I think some projects will be moving to httpx and it's best to minimize our footprint. We only use it for the single download of the core agent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@
 - Move from **requests** to underlying **urllib3** for downloads to reduce
   dependency footprint
   ([PR #480](https://github.com/scoutapp/scout_apm_python/pull/480)).
-- Retry downloading the core agent a few times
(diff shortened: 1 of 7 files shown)


adamchainz:
Error is in a new flaky looking test:

```
=================================== FAILURES ===================================
___________________________ test_report_app_metadata ___________________________
send = <MagicMock name='send' id='140358827793152'>
    @mock.patch("scout_apm.core.socket.CoreAgentSocketThread.send")
    def test_report_app_metadata(send):
        report_app_metadata()


>       assert send.call_count == 1
E       AssertionError: assert 3 == 1
E        +  where 3 = <MagicMock name='send' id='140358827793152'>.call_count
tests/unit/core/test_metadata.py:17: AssertionError
```

It's only happening on Python 3.5, randomly on one of the environments. Will investigate.

PR #489, opened 2020-03-02
Make objtrace C extension optional

### LINKED ISSUE #488: Installation seems to be broken on python3.6.4
Screen Shot 2020-02-26 at 12 31 00 PM

adamchainz:
Fixes #488.

I tested this by deliberately introducing a syntax error into `_objtrace.c`. Without `optional=True`, installation in a virtual environment failed; with it, it passes, and without any error message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Pending
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


adamchainz:
https://docs.python.org/3/distutils/apiref.html#distutils.core.Extension

PR #491, opened 2020-03-02
Implement max span limit

### LINKED ISSUE #472: Total span limit
For *really* long running background jobs it'd be possible to have a tracked request with so many spans it causes memory problems. Following #471 we should probably have a (very high) total cap on the number of captured spans.

adamchainz:
Fixes #472.

### DIFF
--- a/src/scout_apm/core/tracked_request.py
+++ b/src/scout_apm/core/tracked_request.py
@@ -32,7 +32,11 @@ class TrackedRequest(object):
         "is_real_request",
         "_memory_start",
         "n_plus_one_tracker",
+        "hit_max",
     )
+
+    # Stop adding new spans at this point, to avoid exhausting memory
+    MAX_COMPLETE_SPANS = 1500
 
     @classmethod
     def instance(cls):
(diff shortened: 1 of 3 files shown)


adamchainz:
Pending review / Slack discussion

cschneid:
That’s what we do in other agents as well. 1500 is our normal number

This change is fine, although make sure you still send the first set of spans, don’t just drop it entirely
(not sure what setting ignore does)

This will likely have a more full-featured solution later, coordinated with the other language agents, with special tags or something that indicates trimming happened.

src/scout_apm/core/tracked_request.py
```diff
@@ -34,6 +34,9 @@ class TrackedRequest(object):
         "n_plus_one_tracker",
     )
 
+    # Stop adding new spans at this point, to avoid exhausting memory
+    MAX_COMPLETE_SPANS = 100000
```
dlanderson: Let's lower this to 1500 to match our Ruby agent.

adamchainz:
We aren't recording or annotating ignored spans, they're dropped when they hit 'stop': https://github.com/scoutapp/scout_apm_python/blob/f69d0343b9005c9c5d78d11b56d7aca26bdb6270/src/scout_apm/core/tracked_request.py#L111-L113

src/scout_apm/core/tracked_request.py
adamchainz: Done

dlanderson:
Totally missed that. Thanks @adamchainz ! 👍

PR #505, opened 2020-03-16
Ignore asgiref triggered asyncio deprecation warning

adamchainz:


adamchainz:
Will be fixed in https://github.com/django/asgiref/pull/151

PR #512, opened 2020-04-07
Prevent core agent output at startup

### LINKED ISSUE #510: Django log levels are ignored by core agent
No matter what I did to the Django log levels, I kept getting this message when running Django management tasks:

```
[2020-04-02T09:36:06][core_agent][INFO] Initializing logger with log level: Info
```

This is super annoying, because this causes programs in my crontab to start sending out lots of e-mails because previously quiet tasks would now start producing output.

After a lot of trial and error, I figured out that you can set `SCOUT_LOG_LEVEL` in Django settings to suppress this message.

This really should be fixed so the agent also honors the Django logging settings, or at least documented under the logging section in the Python documentation. It's really non-obvious that this the agent is a separate thing which ignores the Django logging settings.

adamchainz:
Fixes #510.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 - Instrument new Elasticsearch methods `get_script_context()` and
   `get_script_languages()`.
   ([PR #507](https://github.com/scoutapp/scout_apm_python/pull/507))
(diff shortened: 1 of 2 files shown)


dlanderson:
👍

PR #518, opened 2020-05-07
Fix MANIFEST.in for check-manifest

adamchainz:
For some reason it has started reporting about the tests not being included. Exclude them.


adamchainz:
Test failures fixed in #519

PR #521, opened 2020-05-07
Fix Travis config warnings

adamchainz:
Fix some warnings seen in "Build config validation" on [a recent build](https://travis-ci.org/github/scoutapp/scout_apm_python/builds/684191778/config):

* root: deprecated key sudo (The key `sudo` has no effect anymore.)
* root: missing os, using the default linux
* services: value redis-server is an alias for redis, using redis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
+os: linux
 dist: xenial
+
 env:
   global:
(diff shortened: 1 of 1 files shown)


adamchainz:
Validation shows a ✅ for this PR.

PR #526, opened 2020-06-04
Don't log Scout key during startup

### LINKED ISSUE #297: Don't log scout key
Currently we log the scout key from configuration and in the register method at startup. Security conscious companies don't want license keys in their log files. We should avoid logging the key, though perhaps a hash or a substring could be useful.

adamchainz:
Fixes #297.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Pending Release
+
+### Fixed
+
+- Don't log the Scout key during startup.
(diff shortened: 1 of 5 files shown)


dlanderson:
Part of why we log the scout key is so we can spot improper configuration (env variables not being picked up, using an example key from help docs, etc). Perhaps we can at least log whether the key matches our key pattern?

adamchainz:
Okay. I checked the PHP integration for the standard, and that doesn't log the key at all.

I could log the first 3 characters, and the length?

dlanderson:
First 3 + length is fine, too

PR #545, opened 2020-07-24
get the user without causing side effects

### LINKED ISSUE #544: scout breaks caching for django
We gave scout a try and looked promising but then we quickly had to disable it on production.

Issue is that scout is adding a `Vary: Cookie` header which breaks caching

This normally happens when some code in django accesses the request.user object. I'm assuming scout is trying to add some meta data and accesses it causing this issue.

We've run into this problem ourselves in the past and the way around is it to look for the internal cached user on the request object. Like this:

```
    # going request.user will generate cookie vary headers, but since
    # we aren't changing the output based on this we want to see the user
    # without adding the header, so look for the lazy user
    if request and hasattr(request, '_cached_user'):
        # noinspection PyProtectedMember
        user = request._cached_user
```

I think if replaced the check for user here

https://github.com/scoutapp/scout_apm_python/blob/master/src/scout_apm/django/middleware.py#L139

oppianmatt:
accessing request.user will cause a side effect that breaks caching

this uses the internal django user

fixes #544

### DIFF
--- a/src/scout_apm/django/middleware.py
+++ b/src/scout_apm/django/middleware.py
@@ -136,8 +136,8 @@ def track_request_view_data(request, tracked_request):
         except Exception:
             pass
 
-    user = getattr(request, "user", None)
-    if user is not None:
+    if request and hasattr(request, '_cached_user'):
+        user = request._cached_user
         try:
             tracked_request.tag("username", user.get_username())
         except Exception:


adamchainz:
Hi thanks for this. I'm working on the issue and extending the test suite to catch this. I'll merge your commit as part of that PR.

adamchainz:
New PR: #546.

oppianmatt:
No worries. Thanks for being so prompt. 

Looking forward to trying scout again in production as it already identified some issues during it's brief run.

PR #548, opened 2020-07-24
Try v2 of GitHub Actions setup-python

adamchainz:


adamchainz:
Same error as reported in https://github.com/actions/setup-python/issues/122

Issue #534, opened 2020-07-18
Support aioredis & aioelasticsearch


yoav-orca:
Hi, we love using scout but we are missing the async support in python, specifically for Redis (aioredis) & ElasticSearch (aioelasticsearch), are there any plans to support them? Will you accept PRs for adding support for them?

adamchainz:
Hi @yoav-orca 

Thanks for the support. We don't have any specific issues tracking aioredis or aioelasticsearch. We'd love to expand support for them though.

If you're interested in instrumenting them you can probably implement something within your own project or a fork of scout-apm. The instruments submodule contains our current instrumentation of libraries: https://github.com/scoutapp/scout_apm_python/tree/master/src/scout_apm/instruments . The main difficulty you'll find with making an implementation within scout-apm is keeping all async code in the `scout_apm.async_` module because we support Python 2 still. However you look into adding instrumentation we'd love to see it and will consider PR's.

Thanks,

Adam

Issue #469, opened 2020-02-05
Support preserving request across asyncio tasks


adamchainz:
Currently an async view in e.g a starlette application that launches and awaits some asyncio tasks will not have them in its trace.

Our current context holder, [asgiref.local](https://github.com/django/asgiref/blob/master/asgiref/local.py), deliberately doesn't restricts context per task. This makes sense since they can be left runnign after the request completes, for example for Starlette's background tasks.

We should provide some way of preserving the tracked request into launched tasks. Need to research state of the art here.

bitdivision:
@adamchainz Are there any plans to solve this issue? As far as I can see this makes instrumenting anything within fastAPI / Starlette non-viable since each coroutine will get a different `TrackedRequest`?

I've implemented a fix by storing the `TrackedRequest` in a context variable and overriding the `scout_apm.api` methods to pull from that context variable, though I believe this would break background tasks. Do you have any ideas on how you would like to solve this in the long-term so as to support background tasks too?

adamchainz:
I'm no longer maintaining Scout, @tim-schilling has taken over.

PR #552, opened 2020-08-05
adding support for aioredis (#534)

yoav-orca:
An initial PR for adding support `aioredis`, ref #534 I have encountered an unfortunate limitation.
When using:

```python
# view code

async def redis_pool():
    # Redis client bound to pool of connections (auto-reconnecting).
    redis = await aioredis.create_redis_pool(
        'redis://localhost')
    await redis.set('my-key', 'value')
    val = await redis.get('my-key')
    print(val)

    # gracefully closing underlying connection
    redis.close()
    await redis.wait_closed()

def index(request):
    loop = get_event_loop()
    loop.run_until_complete(redis_pool())

```

Looking at the trace it seems that scout doesn't recognize that the task is coming from the same thread (the trace is missing the aioredis) segment.

This has already an issue https://github.com/scoutapp/scout_apm_python/issues/469

When using `async_to_sync` it works

### DIFF
--- a/src/scout_apm/async_/instruments/redis.py
+++ b/src/scout_apm/async_/instruments/redis.py
@@ -1,0 +1,83 @@
+# coding=utf-8
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import logging
+
+import wrapt
+
+from scout_apm.core.tracked_request import TrackedRequest
+
+try:
+    import aioredis
+except ImportError:
+    aioredis = None
+else:
+    from aioredis import Redis as AioRedis
+    from aioredis.commands import Pipeline as AioPipeline
+
+logger = logging.getLogger(__name__)
+
+
+have_patched_aioredis_execute = False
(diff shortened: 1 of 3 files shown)


yoav-orca:
@adamchainz what do you think?

adamchainz:
Hi @yoav-orca 

Sorry I didn't respond until now - I went on holiday and your PR was one of many things to respond to.

Thanks for making this. It has certainly exposed the places in aioredis to instrument.

Yes there's still the problem from #469. This needs more research to overcome before we can merge anything.

One thing about your PR - our instruments are *per library* rather than *per "technology"*, so this should be an "aioredis" instrument. Rather than push to your master branch, I've done this in #562 (maintaining your orignal authorship). As such I'm going to close this PR and continue working on the other PR.

We have some other higher priority work around making the launching the core agent more stable, which I'll be addressing first, but I should find time in the next few weeks to revisit aioredis.

Thanks,

Adam

PR #562, opened 2020-09-01
Instrument aioredis

adamchainz:
Rebuild of #552 with tracking in an aioredis-specific instrument submodule.


adamchainz:
This continues to suffer from the problem in #469 - instrumentation won't be preserved for async tasks triggered underneath the main one. We should solve it before merging this.

tim-schilling:
There are changes occurring for pipelines in aioredis v2 which is in alpha currently. It appears that the API will be different so we may want to hold off on this until v2 is finalized.

PR #563, opened 2020-09-01
Fixed validation check on length of key

adamchainz:


cschneid:
Double checked the APM code and it looks like you're correct.

PR #566, opened 2020-09-01
Test with Django 3.1

adamchainz:


adamchainz:
Depends on https://github.com/encode/django-rest-framework/pull/7512 and probably https://github.com/encode/django-rest-framework/pull/7513

PR #587, opened 2020-10-06
Capture full URL in urllib3 spans

adamchainz:
Spotted whilst debugging a customer's traces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Pending
+
+### Fixed
+
+- Capture the full URL in `urllib3` spans.
+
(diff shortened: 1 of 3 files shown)


adamchainz:
I think there's still a bug here where we sometimes capture the wrong hostname, due to a connection pool getting reused for different hosts. It seems `requests` does this as an optimization.

tim-schilling:
I've dug into this and tried recreating it via requests and urllib3. I exceeded the pool size limits and increased the count to see if it was an odd bug exposed by repetition. I was unable to reproduce the bug. Looking at the code, I'm not sure how it would exist since we generate the absolute url from the wrapped connection instance.

If CI passes, I'm inclined to merge this.

Issue #589, opened 2020-10-19
Add typing information


wagdez:
We currently ignore typing of scout_apm in our project
import scout_apm.api  # type: ignore
it would be nice to have typing information

adamchainz:
Hi @wagdez !

I agree, it would be nice to ship type hints for the public API. We're a little constrained on that since we still support Python 2.7, since so many customers use it. Although it's possible to ship type stub files alongside Python 2.7 code, there's at least the unicode/str mess to address.

To be honest we'd need to do some more research to figure out the approach there - personally I've only worked with type hints in application code bases, from Python 3.7+. If you can point to any resources on maintaining type hints on Python 2.7+ projects that'd be useful.

Thanks,

Adam

PR #593, opened 2020-10-19
added partial typing for apis

yoav-orca:
For addressing #589


adamchainz:
Thank you. We'll want to run `mypy` to get this checked on CI - wrong type hints are worse than no type hints. I can take a look at this.

PR #598, opened 2021-01-20
Support hexidecimal amazon time strings.

tim-schilling:
Occassionally Amazon will send the time string in hexidecimal format.
It's not parseable in the format, but they do indicate it will be 8
characters long. It should be reasonable that decimal format time strings
will be longer than that.

### DIFF
--- a/src/scout_apm/core/web_requests.py
+++ b/src/scout_apm/core/web_requests.py
@@ -137,8 +137,10 @@ def track_amazon_request_queue_time(header_value, tracked_request):
     if not first_char.isdigit():
         return False
 
+    # If the timestamp is 8 characters long, it's in hexadecimal format.
+    base = 16 if len(timestamp_str) == 8 else 10
     try:
-        start_timestamp_ns = int(timestamp_str) * 1000000000.0
+        start_timestamp_ns = int(timestamp_str, base) * 1000000000.0
     except ValueError:
         return False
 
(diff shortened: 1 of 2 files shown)


src/scout_apm/core/web_requests.py
```diff
@@ -137,8 +137,10 @@ def track_amazon_request_queue_time(header_value, tracked_request):
     if not first_char.isdigit():
         return False
 
+    # If the timestamp is 8 characters long, it's in hexadecimal format.
+    base = 16 if len(timestamp_str) == 8 else 10
```
adamchainz, line 141: until what date/time will this assumption hold? hex timestamps will spillover into nine digits after 'ffffffff'

tim-schilling: The minimum date of the binary epoch timestamp of 10 digits is 2001/9/9. So we could potentially change this to be 8 or 9 digits, convert to hex, but the Amazon docs explicitly state the hexadecimal value will be rendered as 8 characters. I'm not sure how you want to manage those risks.

The [Elastic Load Balancer docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html) state the time is formated in the number of seconds, but does not list if it's in hexadecimal. The examples show 8 characters, but do not contain a-f. On the other hand, [the XRay docs](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html) show more detailed information about the trace id that does include the hexadecimal notice.

dlanderson: This seems like a reasonable fix. 👍

PR #617, opened 2021-02-09
Upgrade GitHub Actions to Ubuntu 20.04

adamchainz:


adamchainz:
We can't do this whilst continuing to support Python 3.4 - the setup-python action does not support it and bails with:

```
Run actions/setup-python@v2
  with:
    python-version: 3.4
    token: ***
  env:
    ELASTICSEARCH_URL: http://localhost:9200/
    MONGODB_URL: mongodb://localhost:27017/
    REDIS_URL: redis://localhost:6379/0
Version 3.4 was not found in the local cache
Error: Version 3.4 with arch x64 not found
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

tim-schilling:
This was handled in #681

PR #639, opened 2021-03-31
Exclude python library paths from the backtrace.

### LINKED ISSUE #514: Clean library/frameworks from backtraces
We should clean all non-user code from backtraces as much as possible.  Here's a heroku app running 2.13.0:

Screenshot 2020-04-13 14 39 56

tim-schilling:
Closes #514


dlanderson (approved):
👍

PR #650, opened 2021-05-27
WIP - Error monitoring

tim-schilling:
This lacks tests, documentation, integration with the server and the design is still subject to change.

TODO:
- [ ] Django instrumentation unit tests
- [ ] ErrorMonitor core unit tests
- [ ] Integration tests
- [ ] Documentation of new settings and error monitoring feature.
- [ ] Batch errors together and send on separate thread from the requests thread.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,7 @@ setup(
         'urllib3[secure] < 1.25 ; python_version < "3.5"',
         'urllib3[secure] < 2 ; python_version >= "3.5"',
         "wrapt>=1.10,<2.0",
+        "requests",
     ],
     keywords="apm performance monitoring development",
     classifiers=[
--- a/src/scout_apm/django/middleware.py
+++ b/src/scout_apm/django/middleware.py
@@ -5,11 +5,14 @@ import sys
 
 import django
 from django.conf import settings
+from django.views.debug import SafeExceptionReporterFilter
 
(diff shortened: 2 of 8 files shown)


src/scout_apm/django/middleware.py
```diff
@@ -22,38 +25,78 @@
...
+    if django_admin_components:
+        view_name = "{}.{}.{}".format(
+            django_admin_components.module,
+            django_admin_components.controller,
+            django_admin_components.action,
+        )
+
+    django_rest_framework_components = _get_django_rest_framework_components(request)
+    if django_rest_framework_components is not None:
+        view_name = "{}.{}".format(
```
tim-schilling, line 42: One thing to investigate is whether we can and do we want to change the operation name here to include the `module`.

setup.py
```diff
@@ -67,6 +67,7 @@
         'urllib3[secure] < 1.25 ; python_version < "3.5"',
         'urllib3[secure] < 2 ; python_version >= "3.5"',
         "wrapt>=1.10,<2.0",
+        "requests",
```
tim-schilling, line 70: If we are trying to limit our dependencies, we can use `urllib3`, but `requests` makes things much cleaner.

tim-schilling:
Using #651 instead

Issue #416, opened 2019-12-09
Derive application_root for more frameworks


adamchainz:
Currently `application_root` is automatically populated for Django, and only when the pseudo-setting `BASE_DIR` is set (this is a convenience setting that [the Django startproject template includes](https://github.com/django/django/blob/stable/2.2.x/django/conf/project_template/project_name/settings.py-tpl#L16) but isn't used directly in the framework).

This variable is used on the dashboard side to derive which code belongs to the current application for linking to that code on GitHub.

It would be useful to derive it for all frameworks. It might need a heuristic approach based on checking a stack trace of what's running, whether there's a virtualenv, etc.

adamchainz:
TIL the Python sysconfig library gives access to paths where third party packages live https://docs.python.org/3/library/sysconfig.html

PR #651, opened 2021-05-27
Error monitoring

tim-schilling:
Monitor errors via got_request_exception signal.

The middleware process_exception would only allow us to track
exceptions thrown in views, that weren't picked up by an
earlier middleware that returned a result. This solution allows
us to track exceptions that occur anywhere in the application.

The downside to this is that it's possible if another receiver
of got_request_exception throws a new exception, our handler
may get the wrong exception with sys.exc_info().

We could limit the scope of that issue by injecting a middleware
that's as far down the list as possible to wrap the view
and catch exceptions there. Then rely on the signal handler
for catching exceptions in the middleware.

Create ErrorServiceThread to buffer error messages.

TODO:

- [x] Django instrumentation unit tests
- [x] ErrorMonitor core unit tests
- [x] Documentation of new settings and error monitoring feature.
- [x] Batch errors together and send on separate thread from the requests thread.
- [x] API for manually tracking errors

### DIFF
--- a/src/scout_apm/core/config.py
+++ b/src/scout_apm/core/config.py
@@ -223,6 +223,10 @@ class Defaults(object):
             "core_agent_version": "v1.3.0",  # can be an exact tag name, or 'latest'
             "disabled_instruments": [],
             "download_url": "https://s3-us-west-1.amazonaws.com/scout-public-downloads/apm_core_agent/release",  # noqa: B950
+            "errors_batch_size": 5,
+            "errors_enabled": False,
+            "errors_ignored_exceptions": (),
+            "errors_host": "https://errors.scoutapm.com",
             "framework": "",
             "framework_version": "",
             "hostname": None,
--- a/src/scout_apm/core/config.py
+++ b/src/scout_apm/core/config.py
@@ -212,7 +212,7 @@ class Defaults(object):
     def __init__(self):
         self.defaults = {
             "app_server": "",
-            "application_root": "",
+            "application_root": os.getcwd(),
             "collect_remote_ip": True,
             "core_agent_dir": "/tmp/scout_apm_core",
             "core_agent_download": True,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## Pending
 
 ### Added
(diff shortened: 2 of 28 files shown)


src/scout_apm/core/config.py
```diff
@@ -212,7 +212,7 @@ class Defaults(object):
     def __init__(self):
         self.defaults = {
             "app_server": "",
-            "application_root": "",
+            "application_root": os.getcwd(),
```
dlanderson, line 215: Is this directly related to errors, or should this be a separate change? What are the potential liabilities for defaulting this to CWD?

dlanderson:
Where does ErrorServiceThread get started? Does it start even if `errors_enabled` config is false?

tim-schilling:
> Where does ErrorServiceThread get started? Does it start even if errors_enabled config is false?

`ErrorServiceThread.ensure_started` is called at the end of `ErrorServiceThread.send`. This keeps the thread from being created until it's needed.

src/scout_apm/core/config.py
tim-schilling: This change could be pulled into a separate PR. Having the noise of the full path in the error stack trace elevated the priority of #416 as the stack trace is a big part of the error monitoring product.

The risks with this is that it will base `application_root` based on how the application is run rather than how the application is structured. If someone runs the app from a parent directory, the directories between the code and where the application is being run will cause additional directories to appear in stack traces within the Scout app.

dlanderson:
@tim-schilling If we've tested and all looks good and you're ready to release let's go ahead 👍

src/scout_apm/core/config.py
tim-schilling: @dlanderson Are you good with the above risk? We could put the working directory as another value, release it and then review the values to see if it would cause any issues.

tim-schilling:
I'm going to push the existing changes out in a version, then merge this as another version.

src/scout_apm/core/config.py
dlanderson: I think it's good - it should be an improvement for anyone not using django where we can use BASE_DIR

PR #670, opened 2021-08-03
Switch to use ContextVars to store TrackedRequest.

tim-schilling:
This better supports when processes use separate OS threads
to execute code such as run_in_executor

This adds the dependency contextvars when using python 3.6


dlanderson (approved):
LGTM! 👍

PR #746, opened 2022-08-10
Deprecation notice about urllib3[secure]

umarfarouk98:
### Description

 pyOpenSSL and urllib3[secure] are deprecated in the upcoming release (1.26.12)
 https://github.com/urllib3/urllib3/issues/2680
 Removed 'urllib3[secure] < 1.25 ; python_version < "3.5"' and  'urllib3[secure] < 2 ; python_version >= "3.5"'

Closes #746

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,8 +65,6 @@ setup(
         'contextvars ; python_version >= "3.6" and python_version < "3.7"',
         'importlib-metadata ; python_version < "3.8"',
         "psutil>=5,<6",
-        'urllib3[secure] < 1.25 ; python_version < "3.5"',
-        'urllib3[secure] < 2 ; python_version >= "3.5"',
         "wrapt>=1.10,<2.0",
     ],
     keywords=["apm", "performance monitoring", "development"],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -65,6 +65,8 @@
         'contextvars ; python_version >= "3.6" and python_version < "3.7"',
         'importlib-metadata ; python_version < "3.8"',
         "psutil>=5,<6",
+        'urllib3 < 1.25 ; python_version < "3.5"',
+        'urllib3 < 2 ; python_version >= "3.5"',
```
tim-schilling, line 69: Is there a reason to limit the version of urllib3 with py3.5+?

umarfarouk98: It was part of your requirements.txt to limit the versionof the urllib3[secure].
i'll just update with a new pull request as the deprecation doesn't have effect on urllib3 1.25.x or earlier

tim-schilling: Thanks for pointing that out. I need to investigate that.

PR #755, opened 2023-06-28
Fix build

tim-schilling:
Accumulating all the changes from pending PRs to see if the build will pass entirely. This PR should be closed, not merged.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,6 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-        - "3.6"
         - "3.7"
         - "3.8"
         - "3.9"
@@ -27,7 +26,7 @@ jobs:
 
     services:
       elasticsearch:
-        image: elasticsearch:8.2.3
+        image: elasticsearch:8.7.0
         ports:
         - 9200:9200
         env:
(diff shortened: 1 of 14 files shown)


lancetarn (approved):

tim-schilling:
@quinnmil @mitchh456 Sorry, I've been struggling to make time to address these open PRs I have. I think the code is sound. The issues with py2.7 and py3.4 are because we don't have GitHub action images for those versions. What are your thoughts?

quinnmil:
@tim-schilling No worries. Our team has decided to drop support for Python versions < 3.8 and ensure that we fully support 3.11. Along with this, we'll be removing some integrations for libraries that are no longer commonly used.

Our data shows that very few clients are still using these older versions/integrations, so we effectively want to slim down this project to simplify future maintenance. We're planning on this being a new major version release so we can instruct users to pin previous versions if they want to retain support for the older python versions/libraries. 

Many of these changes/removals have been done in the [dev](https://github.com/scoutapp/scout_apm_python/tree/dev) branch and we had made many of the same changes you did here before finding this PR and related branches.

I will discuss with @mitchh456 next week on if we'd like to merge some/all of your PRs, but since we're no longer concerned about supporting these old Python and library versions, the entire test suite (and all those tox dependencies) can be hugely simplified. 

So nothing actionable right now, just wanted to give you an update. Thanks for all your work on this project!

PR #760, opened 2023-08-09
757 remove support for unused frameworks - pt 1

quinnmil:
## Changes
- Removes support for CherryPy, Pyramid, and Django versions < 3.2
- Removes the agent implementation for these frameworks as well as their integrations tests and related testing requirement imports.


tox.ini
```diff
@@ -18,34 +18,16 @@ deps =
     attrs ; python_version > "3.4"
     attrs<21.1.0 ; python_version <= "3.4"
     bottle
-    cherrypy ; python_version > "3.5"
-    cherrypy<18.6.1 ; python_version <= "3.5"
-    celery!=4.4.4  # https://github.com/celery/celery/issues/6153
```
mitchh456, line 23: I don't think celery should be removed here

mitchh456 (approved):

PR #762, opened 2023-08-10
remove eol python versions

mitchh456:


setup.py
```diff
@@ -61,28 +59,16 @@
...
         "Framework :: Django",
-        "Framework :: Django :: 1.8",
-        "Framework :: Django :: 1.9",
-        "Framework :: Django :: 1.10",
-        "Framework :: Django :: 1.11",
-        "Framework :: Django :: 2.0",
-        "Framework :: Django :: 2.1",
-        "Framework :: Django :: 2.2",
-        "Framework :: Django :: 3.0",
-        "Framework :: Django :: 3.1",
```
quinnmil, line 85: I think we also want to delete line 75 
```
"Framework :: Pyramid",
```
while were here.

PR #765, opened 2023-08-14
764 update agent to fully support python 311

mitchh456:
This PR adds py311 to tox environments and fixes a couple small issues.
## Changes
- Test fixture to mock `ErrorServiceThread.send`. For some reason in 3.11, the return type of the wrapper wasn't callable, so invoking `ErrorServiceThread.send` threw `AttributeError: 'method' object has no attribute __get__`. This fix is backwards compatible and works with previous python versions.
- Similar to <py35, in py311 you can't force memory reallocations, so the conditional skip for that test was updated to skip >=3.11.


quinnmil (approved):
Good stuff, and nice work on that odd `ErrorServiceThread.send` error.

PR #767, opened 2023-08-17
766 fix sqlalchemy integration tests

quinnmil:
## Changes

- Reworks and simplifies flask-sqlalchemy instrumentation. 
- flask-sqlalchemy v3.0.0+ requires the plugin to be scoped to an active flask app. ([changelog](https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/changes/#version-3-0-0)) ~~This requires rewriting our integration tests but allows us to hugely simplify how we can get the db object. This turns the `flask-sqlalchemy` instrumentation into a one line wrapper around the `sqlalchemy` implementation.~~
UPDATAE: We are instead pulling in changes originally authored by @tim-schilling which retain support for older versions of flask-sqlalchemy and instead use a manual context in the integration tests and wraps `make_engine` for flask-sqlalchemy version >= 3.0.0
- Updates integration tests for `flask-sqlalchemy` and `sqlalchemy` to use the `text` method, now required by sqlalchemy to use raw SQL strings.


tests/integration/test_flask_sqlalchemy.py
```diff
@@ -39,9 +42,10 @@ def test_sqlalchemy(tracked_requests):
 
     assert response.status_int == 200
     assert response.text == "Hello from the DB!"
-    assert len(tracked_requests) == 1
+    # assert len(tracked_requests) == 1
```
mitchh456, line 45: Is this check failing? looks like for the rest of the asserts to pass `tracked_requests` is at least not empty

quinnmil: This was commented out for some testing I was doing, forgot to remove. Good catch.

quinnmil:
Code merged or duplicated in https://github.com/scoutapp/scout_apm_python/pull/769

PR #771, opened 2023-09-07
remove in-code python version checks

mitchh456:
As we only support Py3.8+ in the coming version of the agent, there is no need for all of these version checks in the code.


quinnmil (approved):
Awesome, hugely simplifying.

PR #773, opened 2023-09-12
Handle newer urllib3

mitchh456:
This PR updates urllib3 instrumentation to handle version >= 2 and removes the limit requirement. 

urllib3[secure] is being deprecated(see [here](https://github.com/urllib3/urllib3/issues/2680)), however we still need `certifi` so is is now specified in the dependencies.


quinnmil (approved):
👍

PaperNathan (approved):
👍

PR #790, opened 2024-07-01
Unpin psutil dependency

tim-schilling:


lancetarn:
Thanks Tim! This is a good reminder that some more cleanup is due. These test failures seem unrelated, we'll have to see what else has shifted here.

DmytroLitvinov:
Came here to say same :)

tim-schilling:
Sounds like this check was due to this change is psutil? https://github.com/giampaolo/psutil/issues/2396

Makes sense. Glad to see this moving forward!

lancetarn (approved):

lancetarn:
Yeah. I doubt it was going to be a problem but it seems harmless to look out for it.
We've been heads-down trying to bring log management features to life and haven't done a great job threading maintenance in between... 😬

PR #800, opened 2024-10-04
Remove use of datetime.utcnow()

### LINKED ISSUE #799: Remove use of datetime.utcnow()
Originally attempted back in #368 but the change was reverted in #400. On Python 3.12, `datetime.utcnow()` now triggers this warning:

> DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).

Time to actually migrate.

adamchainz:
Fixes #799.

This needs further testing to check whether the issue from #400 is fixed in the core agent or if it needs handling here, by changing the formatting to remove timezone info.

I no longer have a working dev setup for this though, so I will leave that testing in your hands. (I used to be the maintainer of this package.)

### DIFF
--- a/src/scout_apm/celery.py
+++ b/src/scout_apm/celery.py
@@ -29,7 +29,9 @@ logger = logging.getLogger(__name__)
 
 def before_task_publish_callback(headers=None, properties=None, **kwargs):
     if "scout_task_start" not in headers:
-        headers["scout_task_start"] = datetime_to_timestamp(dt.datetime.utcnow())
+        headers["scout_task_start"] = datetime_to_timestamp(
+            dt.datetime.now(dt.timezone.utc)
+        )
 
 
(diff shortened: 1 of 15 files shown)


lancetarn:
Thank you for this, Adam. And for all your work here and with Django. We'll run it through the test apps... which you have also contributed. This is one of those changes that my soul knows is good for python but is not exactly going to be painless. We will schedule a look at the core agent to check out options.

lancetarn:
Work is in-flight to handle multiple timestamp formats in the Core Agent.

lancetarn:
I think the Core Agent can actually handle offset information in an expected format as-is. I've collected these changes and made some tiny tweaks in #801 and will bring them home there. Thank you so much again for the major effort here!

PR #803, opened 2025-01-03
Parse new configuration options

quinnmil:
## Changes
- Adds new Config `known_keys` and converters to support the new sampling configuration options. 
- Also used copilot to generate types for all these methods because why not? 
- Pins `urllib3` to `~2.2.0` due to [this issue](https://github.com/gabrielfalcao/HTTPretty/issues/484)


src/scout_apm/core/config.py
```diff
@@ -223,34 +229,43 @@ def __init__(self):
...
             "hostname": None,
+            "ignore": [],
+            "ignore_endpoints": [],
+            "ignore_jobs": [],
             "key": "",
             "log_payload_content": False,
             "monitor": False,
             "name": "Python App",
             "revision_sha": self._git_revision_sha(),
+            "sample_rate": 1.0,
```
mitchh456, line 251: I think `100` to match the rest

quinnmil: yup, good catch.

mitchh456 (approved):

PR #805, opened 2025-01-03
Add sampling logic

quinnmil:
## Changes
- Adds `Sampler` class which handles: 
- Loading sampling configuration
- Finding the matching job/controller pattern for a given operation name. Update: This has been simplified to just use a `startswith` to be consistent with how the current `ignore_path` logic works. 
- get sampling rate for matched job/controller, make sampling decision based on priority order. 
- Updates `Config` module to accept new configuration options: `sample_endpoint_rate` and `sample_job_rate` which work as a endpoint/job catchall, with the fallback value still being the global `sample_rate`.


src/scout_apm/core/sampler.py
```diff
@@ -0,0 +1,157 @@
...
+        if name in patterns:
+            return name
+
+        # Then check for wildcard patterns, prioritizing longest match
+        matching_pattern = None
+        longest_match = 0
+
+        wildcard_patterns = [p for p in patterns if "*" in p]
+        for pattern in wildcard_patterns:
+            if pattern.endswith("*"):
```
mitchh456, line 77: Not that this is wrong, but existing ignore logic simply matches prefix. https://github.com/scoutapp/scout_apm_python/blob/c9af2379a2791104cf08f841f80edce3c097cdbe/src/scout_apm/core/web_requests.py#L96
I don't think handling wildcard is necessary

quinnmil: So the thinking here was that a user could want to _sample_`/foo/bar*` differently than they would want to sample `/foo/baz/*` and maybe only completely ignore `foo/biz/*` 

The ignore_path will just outright ignore, but I was under the impression we wanted to support distinct sampling levels with wildcards. Hence my question in your PR as well. We should discuss further.

src/scout_apm/core/sampler.py
```diff
@@ -0,0 +1,150 @@
...
+
+        Returns:
+            Boolean indicating if any sampling is enabled
+        """
+        return (
+            self.sample_rate < 100
+            or self.sample_endpoints
+            or self.sample_jobs
+            or self.ignore_endpoints
+            or self.ignore_jobs
```
mitchh456, line 52: I think `endpoint_sample_rate` and `job_sample_rate` should be checked here as well

src/scout_apm/core/sampler.py
```diff
@@ -0,0 +1,150 @@
...
+        Returns:
+            Integer between 0 and 100 representing sample rate
+        """
+        op_type, name = self._get_operation_type_and_name(operation)
+        if not op_type or not name:
+            return self.sample_rate  # Fall back to global rate for unknown operations
+
+        if op_type == "endpoint":
+            if name in self.ignore_endpoints:
+                return 0
```
mitchh456, line 115: Should individual sample should be before ignore? This is something we talked about last week

quinnmil: I think the ignore should be first because if `/users/test` is ignored, and we have `/users:50` we want to ignore the former instead of sampling via the matched prefix. Any mistakes in this thinking?

mitchh456: what if `/foo/bar:50` but `/foo` is ignored?

quinnmil: > what if /foo/bar:50 but /foo is ignored?
I think this would be fine. 
edit: I added a test to confirm so. 

So I think since _this_ ignore check is just doing a `if name in self.ignore_endpoints:` so: it's looking for an exact match. for _this_ code, you can sample the full length and the ignore will only happen if the endpoint is only eh shorter one. The existing ignore code which you'd previously linked to here: [scout_apm_python/src/scout_apm/core/web_requests.py](https://github.com/scoutapp/scout_apm_python/blob/c9af2379a2791104cf08f841f80edce3c097cdbe/src/scout_apm/core/web_requests.py#L96)

will ignore based on prefix. In the next PR #806, I think I should just remove the check for `self.is_ignored` entirely, and rely on this sampler class to do all the sampling+ignoring, which I'd been considering in the first place.

quinnmil: Okay, so in thinking more about my last point/idea there, I realized the true issue was how we could keep existing functionality of the `ignore` configuration (ignoring via prefix) while still allowing fully specified paths to override this. Had to do a somewhat significant refactoring/reworking of this method and a couple others. Now, the prioritization order is more specific:
```
        Priority order (highest to lowest):
        1. Exact matches in sample_endpoints/sample_jobs
        2. Exact matches in ignore lists (returns 0)
        3. Prefix matches in sample_endpoints/sample_jobs
        4. Legacy ignore patterns (returns 0)
        5. Request-level ignore (returns 0)
        6. Operation-specific default rate
        7. Global sample rate
  ```

File: src/scout_apm/core/sampler.py [hunk 4]
```diff
@@ -0,0 +1,176 @@
...
+            or self.ignore_jobs
+            or self.endpoint_sample_rate is not None
+            or self.job_sample_rate is not None
+        )
+
+    def _find_exact_match(
+        self, name: str, patterns: Dict[str, float]
+    ) -> Optional[float]:
+        """
+        Finds the exact sample rate for a given operation name.
```
mitchh456, line 62: this method name is confusing

src/scout_apm/core/sampler.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+
+        # Check for exact match in sampling patterns
+        exact_rate = self._find_exact_match(name, patterns)
+        if exact_rate is not None:
+            return exact_rate
+
+        # Check for exact endpoint/job ignores
+        if name in ignored_set:
+            return 0
```
mitchh456, line 137: Ignoring shouldn't be done by exact match for endpoints. It should also be using prefix.

src/scout_apm/core/sampler.py
mitchh456: Okay I understand now, but I still don't see how it's all necessary. The exact match checks, and the "longest match" thing seems like extra steps. And it then means other work to support the old `ignore` independently. 

With something like:
 ```
self.ignore_endpoints = set(config.value("ignore_endpoints") or config.value("ignore"))
```
then checking for match by prefix in self.sample_endpoints, then match by prefix in self.ignore_endpoints, all the same boxes get checked. Your example `/users/test/` and `/users:50` can be solved with the configuration itself.

Comment (quinnmil) [reply src/scout_apm/core/sampler.py]: Okay, core of the confusion here was that I hadn't been considering relying on a sampling configuration like `foo/bar:0, foo:50` instead of only supporting `ignore/ignore_endpoints`. We _can_ still suport both of these cases: 
```
sample: /foo --- ignore: /foo/bar
and
2. sample: /foo/bar --- ignore: /foo
```
Specifically, we allow the (1) option by using a configuration like the one shown above. This lets us simplify the logic significantly.

mitchh456 (approved):

PR #812, opened 2025-05-06
Check `ignore` value for legacy configuration

quinnmil:
## Changes
- Fixes a bug where if the only sampling configuration set was the `ignore` value, we weren't checking for this in the `any_sampling` check. 
- Adds a test for this case.


quinnmil:
Realized that the changes here are redundant. This has already been supported via this: 
```python
        self.ignore_endpoints = set(
            config.value("ignore_endpoints") + config.value("ignore")
        )
```
So we're removing extra logic, but leaving in the new test.

